Issue #49, opened 2017-10-11
[Q] How many profiles are actually supported?


journey2w:
Issued by @MariLB.

----

How many profiles are actually supported at the moment?
Some topics say 3, others say 4, and some say 3 and 4 within the same topic. This is quite confusing.
docs\what-is-tizen\overview\tizen.md:
docs/open-source-tizen/about/tizen-open-source-overview.md
docs/open-source-tizen/compliance/compliance-program.md
docs/open-source-tizen/about/git-repo-structure.md
Trademark symbolds have not been fixed yet, since we do not know how to add them to Markdown. They will be handled later.

Are .jpg images allowed, or should they be converted to .png?
Also, is it OK to have images with widths larger than 850 pixels?
Images have not been fixed yet.

There are lots of details about things happening in (beginning of) 2017. We’re almost at the end now. Have these things happened?

docs/open-source-tizen/tizen-rt/rt-goal.md
docs/open-source-tizen/tizen-rt/tizen-rt-overview.md
The PDFs in the following folder all seem to be corrupted:
docs\open-source-tizen\compliance\media\
docs\open-source-tizen\porting\multimedia.md and docs/open-source-tizen/porting/telephony.md:
Several tables in these topics have multi-line code snippets in table cells, which cannot be done in markdown. We have not fixed them yet, since it is unclear how to implement this in Markdown.

journey2w:
1. How many profiles are actually supported at the moment?
Some topics say 3, others say 4, and some say 3 and 4 within the same topic. This is quite confusing.
docs\what-is-tizen\overview\tizen.md:
docs/open-source-tizen/about/tizen-open-source-overview.md
docs/open-source-tizen/compliance/compliance-program.md
docs/open-source-tizen/about/git-repo-structure.md

=> The supported profiles depend on version.
Tizen 2.3 and lower: Mobile, Wearable, IVI
Tizen 2.3.1: Mobile, Wearable
Tizen 2.3.2: Wearable
Tizen 2.4: Mobile, Wearable
Tizen 3.0 and higher: Mobile, Wearable, TV

It has to be different from which docs. If you dont' know, don't change it.
For instance, compliance program certificates Mobile, Wearable, and TV profiles according to version.
Git repo has to have a git history and the latest updates.

2. Trademark symbolds have not been fixed yet, since we do not know how to add them to Markdown. They will be handled later.
=> fixed PR #52  
[...]

PR #48, opened 2017-10-06
Edit source.tizen.org content

MariLB:
Performed basic editing, fixing language and subsection structures.
Improved usability and readability.

PR #53, opened 2017-10-13
Fix corrupted pdf files and remove codes in tables.

journey2w:
Related PR: #48 
Related issues : #49


l10nbridge:
The tizen_branding_request_form_for_tizen_mobile_profile_v1.0_1.pdf file still has issues, it shows up with a size of 0 and does not open.
The other PDF files are now OK.

PR #63, opened 2017-10-25
Tizen IoT Setup Wizard Binaries

siddarthkm:
- Adding Tizen IoT Setup Wizard v0.4.1-preview binaries as parts
 - combine.sh script provided to combine parts to original jar file
 - Download all files in iot-preview/tools/setup_wizard into a single directory and execute ./combine.sh in a bash shell

### DIFF
--- a/iot-preview/tools/setup_wizard/README.md
+++ b/iot-preview/tools/setup_wizard/README.md
@@ -1,0 +1,16 @@
+Tizen IoT Setup Wizard
+======================
+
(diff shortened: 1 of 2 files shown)


jiehwan:
how about to create README.md to explain the way how to combine compressed parts into jar file.

journey2w:
This github is not officially announced for Tizen docs. I said you can use this repo for temporary storage of a jar file because it is about 50MB.
However, I'm sorry to tell you that the size is too big and the usage is not suitable for this repo.

PR #69, opened 2017-11-09
Convert guides/natvie and guides/web to Markdown

junmoy:


journey2w:
guides > we & guides / web files were converted.
We have to check links after pushing all docs.

PR #139, opened 2018-01-26
Add desc about watch app background time_tick

gotoextreme:
Watch application's time tick can be called in background but document do not contain about this inforamtion.


mijin80lee:
TD Sync Completed.

PR #147, opened 2018-01-30
Match URL paths in native guides /efl part II

s-k-song:
- Match URL paths in /native/guides/ui/efl part II (from component-plug to key-grab)
- Move "Dependencies" to the end or insert missing ones
- Delete spaces in table header
- Re-arrange lists, table, code blocks for proper display


s-k-song:
Removed csharp in the code blocks.

Issue #176, opened 2018-02-12
URL link error in /application/vstools


ck11a:
[/tools/certificate-manager.md]
Web

PR #180, opened 2018-02-13
Fix URL error in vstools

s-k-song:
For the Issue #176, the URL is fixed.


journey2w:
Fixed on PR #815 and #187.

PR #202, opened 2018-02-27
Update native app-control guide

hjhun:
Updates guide for access control with privileges

Signed-off-by: Hwankyu Jhun <h.jhun@samsung.com>


hjhun:
I'll prepare a new patch.

PR #257, opened 2018-04-27
[Filesystem] Amend the remark about proper escaping in the FileFilter

pwasowski2:
The remark about escaping '%' character in FileFilter description did
not take into account the fact, that '' escape character needs itself
to be escaped from the string with another backslash.
The description was updated.

Signed-off-by: Pawel Wasowski <p.wasowski2@partner.samsung.com>


mijin85cho:
The Web API Reference for 2.4, 3.0 and 4.0 are uploaded on Tizen Developer Site.

PR #279, opened 2018-05-23
Add feature key for wifi-direct tethering

SeonahMoon:
network.tethering.wifi.direct feature has been added since tizen 4.0

[related ACR]
http://suprem.sec.samsung.net/jira/browse/ACR-967


mijin80lee:
TD Sync Completed.

PR #316, opened 2018-07-10
Fix link path to immutable index

lu-pik:
Signed-off-by: Lukasz Pik <lu.pik@samsung.com>

### Change Description ###

Fix link to wearable version of immutable index.


mijin80lee:
TD Sync Completed.(for Tizen_5.0_prepare)

PR #323, opened 2018-07-17
[Player] Fix link of media stream api

eunhaechoi:
- Fix media stream link in Player module.


mijin80lee:
TD Sync Completed.(for Tizen_5.0_prepare)

PR #381, opened 2018-09-06
[Dotnet/Sample/SystemInfo] Add detailed description for systeminfo applications

hj-na-park:
Add detailed description to distingish similar applications.


mijin80lee:
TD Sync Completed.

PR #388, opened 2018-09-13
[Native/guide] Fix Privacy related prermissions dependency in docs

xerrni:
Signed-off-by: Ernest Borowski <e.borowski@partner.samsung.com>

### Change Description ###

Fixed dependency of Privacy privileges.
PrivacyPrivilegeManager was introduced in tizen 4.0, but inside docs there is tizen 2.3.x.
This patch should be cherry-picked to all branches.


mijin85cho:
Thanks for the update.

PR #402, opened 2018-09-28
Add "Note" in case of build-config absense

soong9:
Add guide in case of not cloning 'scm/meta/build-config' git.
If 'scm/meta/build-config' git is not cloned in working directory, every line which starts with "buildconf=./scm/meta/build-config" should be removed

### Change Description ###

Describe your changes here.

For instance,
 -  ...

### Bugs Fixed ###

Provide links to bugs here

For instance,

 - Issue #265

### API Changes ###

For instance,

 - ACR: ACR-1120

### DIFF
--- a/docs/open-source-project/developing/setting-up.md
+++ b/docs/open-source-project/developing/setting-up.md
@@ -426,6 +426,10 @@ url = http://download.tizen.org/releases/daily/tizen/3.0-ivi/latest/repos/arm/de
 >
 > The file contains the GBS configuration for all profiles and repositories in Tizen version 3.0, Tizen version 4.0 and Tizen version 5.0. In the near future in new GBS versions, the above default configuration file (`~/.gbs.conf`) is automatically installed when GBS is installed.
 
+> **Note**
+>
+> If "scm/meta/build-config" git is not cloned in your working directory, every line which starts with "buildconf=./scm/meta/build-config" should be removed.
+
 The default profile used in GBS is specified in the `[general]` section:
 
 ```


mijin85cho:
Hello,

The notice is located after the guide. How about giving the note before the user encounter "buildconf=~~" & build error?

For example moving the note to line 146.

docs/open-source-project/developing/setting-up.md
```diff
@@ -426,6 +426,10 @@ url = http://download.tizen.org/releases/daily/tizen/3.0-ivi/latest/repos/arm/de
 >
 > The file contains the GBS configuration for all profiles and repositories in Tizen version 3.0, Tizen version 4.0 and Tizen version 5.0. In the near future in new GBS versions, the above default configuration file (`~/.gbs.conf`) is automatically installed when GBS is installed.
 
+> **Note**
+>
+> If "scm/meta/build-config" git is not cloned in your working directory, every line which starts with "buildconf=./scm/meta/build-config" should be removed.
+
 The default profile used in GBS is specified in the `[general]` section:
```
divyk: Suggestion:

This whole structure can be re-written as:

image

If you think, it is not breaking the flow, please accept this. Thanks

mijin85cho: Hello, @divyk 
To my understand, in this fix, 3 notes and the code block in the middle are merged into one note.
But unfortunately, I don't see the advantage of merging these into **a note**.

the 1st note is additional description for the Default GBS Configuration File.
The 2nd note is about the preventing build error.
The 3rd note is about the GBS build parameter, which is added to the default profile description.

Every note has different subjects, and separated according to their subject.
Would you let me know the purpose of merging these different subjects into one note?

divyk: @mijin85cho , as per my limited knowledge I made this suggestion. Please correct me if I am wrong.

- The default profile.... and The default GBS.... are not under "Note". Those are standalone statements.

- I moved all the text above the code, because all these descriptions are in relation with the code below.

- I moved GBS... above, because it is talking about the code. and it would be better to read this first and then see the code.

This I wrote as per my limited knowledge, please correct me if I am wrong. Thanks in advance.

mijin85cho: @divyk So, you are suggesting to get "the file contains the GBS..." note and "If "scm/meta/build-config" git..." note together, converting "The default GBS build ..." note into a standalone statement.
Also moving all these to above of the sample sample code. 

If so, I think your suggestion is acceptable.

@soong9, would you please check divyk's comment?

mijin85cho:
@soong9, would you please check divyk's comment?

docs/open-source-project/developing/setting-up.md
soong9: @mijin85cho 
I think it's better to keep current order of statement, because I think it's better to show default setting first, and then show notification later.
If notification is shown first, user can be confused the word inside default configuration.

PR #507, opened 2019-01-02
Modify the thumbnail size instead of returning the error

clockr63:
### Change Description ###

If the width and height of the thumbnail to be generated exceeds the original resolution,
the value changes to the original resolution instead of returning the error.

### API Changes ###

 - ACR: ACR-1327


huytaduc:
checkout the webpage https://docs.stg.tizen.org/staging/507

huytaduc:
checkout the webpage https://docs.stg.tizen.org/staging/507

mijin80lee:
TD Sync Completed.(reviewing state)

PR #512, opened 2019-01-07
Change the root of menu tree for Application

journey2w:
### Change Description ###

Docs site is documents for developing Tizen application. 
Updates of documents for Design and Distribute doesn't happen frequently so Design and Distribute menus are removed from the menu tree, but their documents are remained in this repo.


huytaduc:
checkout the webpage https://docs.stg.tizen.org/staging/512

PR #531, opened 2019-01-16
[3.0] [Common] Added missing information about remove listener behaviour

pkosko:
Some listener related API (remove*Listener) has missing description about the behaviour
when the listener was not registered yet. This commits adds such information to every function.


huytaduc:
checkout the webpage https://docs.stg.tizen.org/staging/531

PR #550, opened 2019-01-30
change form of target blank to fix it

haivutuan93:


haivutuan93:
I'll support style with format: {:target="_blank"}
So, i closed this pull request

PR #556, opened 2019-02-12
Update applications doc

hjhun:
The note is added about the app_get_display_state() function.

Signed-off-by: Hwankyu Jhun <h.jhun@samsung.com>

### Change Description ###

Add the note about the app_get_display_state() function.

### API Changes ###

 - ACR: ACR-1310

### DIFF
--- a/docs/application/native/guides/applications/overview.md
+++ b/docs/application/native/guides/applications/overview.md
@@ -44,9 +44,9 @@ When your application becomes visible again, the `app_resume_cb()` callback is i
 - All applications on top of your application in the window stack finish.
 - An alarm is triggered for your application, bringing it to the front and hiding other applications.
 
-	> **Note**
-    >
-	> Since Tizen 5.5, you can get the display state by using the  app_get_display_state()  function when the  app_resume_cb()  callback or the  app_pause_cb() callback is invoked.
+  > **Note**
+  >
+  > Since Tizen 5.5, you can get the display state by using the  app_get_display_state()  function when the  app_resume_cb()  callback or the  app_pause_cb() callback is invoked.
 
 When your application starts exiting, the `app_terminate_cb()` callback is invoked. Your application can start the termination process, when:
 
--- a/docs/application/native/guides/applications/overview.md
+++ b/docs/application/native/guides/applications/overview.md
@@ -44,6 +44,10 @@ When your application becomes visible again, the `app_resume_cb()` callback is i
 - All applications on top of your application in the window stack finish.
 - An alarm is triggered for your application, bringing it to the front and hiding other applications.
 
+        > **Note**
+    >
+        > Since Tizen 5.5, you can get the display state by using the  app_get_display_state()  function when the  app_resume_cb()  callback or the  app_pause_cb() callback is invoked.
+
 When your application starts exiting, the `app_terminate_cb()` callback is invoked. Your application can start the termination process, when:
 
 - Your application itself requests to exit by calling the `ui_app_exit()` or `service_app_exit()` function to terminate the event loop.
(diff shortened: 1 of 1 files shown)


huytaduc:
checkout the webpage https://docs.stg.tizen.org/staging/556

huytaduc:
checkout the webpage https://docs.stg.tizen.org/staging/556

huytaduc:
checkout the webpage https://docs.stg.tizen.org/staging/556

huytaduc:
checkout the webpage https://docs.stg.tizen.org/staging/556

docs/application/native/guides/applications/overview.md
```diff
@@ -44,6 +44,10 @@ When your application becomes visible again, the `app_resume_cb()` callback is i
 - All applications on top of your application in the window stack finish.
 - An alarm is triggered for your application, bringing it to the front and hiding other applications.
 
+  > **Note**
```
Reni-90: Suggestion:
The **Note** is for `app_resume_cb()` and `app_pause_cb()`. The **Note** should also be general. Here, it is aligned with the bullet list.
The alignment should be:

line47

Please verify and fix the alignment.

docs/application/native/guides/applications/overview.md
```diff
@@ -44,6 +44,10 @@ When your application becomes visible again, the `app_resume_cb()` callback is i
...
+  >
+  > Since Tizen 5.5, you can get the display state by using the  app_get_display_state()  function when the  app_resume_cb()  callback or the  app_pause_cb() callback is invoked.
```
Reni-90: Suggestion:
Usage of "since" is not proper as Tizen 5.5 is the current version.

Rephrase as:
From Tizen 5.5, you can get the display state using `app_get_display_state()` when the `app_resume_cb()` callback or the `app_pause_cb()` callback is invoked.

journey2w: @Reni-90 
"From Tizen 0.0, ..." is correct but we usually use "Since Tizen 00.00, ...". You should've referred the guides of Tizen other modules.
It would be good not to change this prase, if there is no grammatical error.

Reni-90: @journey2w 

Actually, I was not convinced with the usage of “since” for current version. Therefore, I researched on the same.
http://pediaa.com/difference-between-since-and-from/ 
https://developers.google.com/style/word-list
https://developer.android.com/topic/libraries/support-library/revisions
https://developer.android.com/ndk/reference/group/sync 

It is proper to use “since for an older version. For example, see https://developer.android.com/topic/libraries/support-library/revisions 
•	AppBarLayout now has 300px of left padding since 26.0.0.
•	If you are not using the Lifecycles library, you don’t need to update from 26.0.2.

Since refers to an action, which has continued up to now. It refers to the passage of time.

I have recommended the change after going through all the above points. 
If you are not convinced then please guide me how to proceed in this situation.

Regarding the consistency, whenever a new version is released, we can use “since for the previous version” and “from for the current version”. If you agree with this. Please let me know what is your recommendation regarding the same.

journey2w: @Reni-90
I usually use "since" with present perfect tense, but the previous outsourcing company sometimes did use present simple or past simple tenses as above sentence. It doesn't seem wrong.

Your comment is reasonable but you seems to misunderstand something. 
The examples that you cited are not matched with your explanation. In "If you are not using the Lifecycles library, you don’t need to update from 26.0.2.", 26.0.2 is not the current version but the previous version. The current version is  26.1.0.

We will release a next version of Tizen someday, so the current version will be the previous version without any updates at that time.

hjhun:
Dear all,
I'm sorry. I've created a new pull request for this.
https://github.com/Samsung/tizen-docs/pull/589

PR #559, opened 2019-02-12
change lnb of api reference use toc common instead toc in js

haivutuan93:


huytaduc:
checkout the webpage https://docs.stg.tizen.org/staging/559

PR #594, opened 2019-02-27
viewSwitcher: Add missing return type

lslachciak:
[Problem] getActiveIndex method signature had return value type missing
[Solution] Add missing return value type

Signed-off-by: Pawel Kaczmarczyk <p.kaczmarczy@samsung.com>


huytaduc:
checkout the webpage https://docs.stg.tizen.org/staging/594

PR #682, opened 2019-04-18
[Content] [5.5] Cleaning code examples from modification of readonly members

pkosko:
Few properties were marked as readonly since 5.5 with below ACR, but the code examples
still existed in API documentation which used to modify the properties now marked as
readonly.
Examples were fixed to use still editable attributes only.

[ACR] http://suprem.sec.samsung.net/jira/browse/TWDAPI-203

### DIFF
--- a/docs/application/web/api/5.5/device_api/mobile/tizen/content.html
+++ b/docs/application/web/api/5.5/device_api/mobile/tizen/content.html
@@ -474,10 +474,10 @@ contain an invalid value.
         </div>
 <div class="example">
 <span class="example"><p>Code example:</p></span><pre name="code" class="examplecode prettyprint">/* Assume the content is a Content object as a result of find method. */
-/* Check if the description is editable, and then set a description. */
-if (content.editableAttributes.indexOf("description") &gt;= 0)
+/* Check if the isFavorite attribute is editable, and then set it as favorite. */
+if (content.editableAttributes.indexOf("isFavorite") &gt;= 0)
 {
-  content.description = "Sample content";
+  content.isFavorite = true;
 }
 tizen.content.update(content);
 </pre>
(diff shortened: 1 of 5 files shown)


docs/application/web/api/5.5/device_api/mobile/tizen/content.html
```diff
@@ -675,6 +675,10 @@ <h4>Methods</h4>
 <p><span class="privilege">Privilege: </span>
  http://tizen.org/privilege/content.read
             </p>
+<p><span class="remark">Remark: </span>
+ Attributes available for <em>Content</em> objects filtering are listed in
+<a href="https://developer.tizen.org/development/guides/web-application/data-storage-and-management/data-filtering-and-sorting#content_filter">Data Filtering and Sorting guide</a>.
```
mijin85cho: Question to the "docs/application/web/guides/data/data-filter.md#content-filter-attributes" file.

In this file we have "Content filter attributes" table. Don't we need to add "isFavorite" to that table?
Also, in this "Table: Content filter attributes", we have "modifedDate" attribute.
I'm wondering whether if "modifiedDate" isn't correct spelling.
Would you please check?

If I understood something wrong, please let me know.

mijin85cho: Hello, @rwalczyna, would you please check my comment? If you think current state is OK to merge, would you please confirm it?

pkosko: @mijin85cho sorry for late response, we were checking your question.
1. I fixed the typo you mentioned.
2. For now, filtering by isFavorite members is not available, thus it cannot be added in the table related to content filtering.

mijin85cho: Thanks for the answer and the update.
@rwalczyna If we have no more issues with this PR, would you please re-approve it?

rwalczyna: For me it also looks okay.

PR #706, opened 2019-05-13
Add a guide about device type key

kichankwon:
### Change Description ###

  - Add a guide about device type key.

### API Changes ###

 - ACR: ACR-1381

### DIFF
--- a/docs/application/native/guides/device/system.md
+++ b/docs/application/native/guides/device/system.md
@@ -70,6 +70,7 @@ To obtain the information, query a feature or system key.
 - [System keys](#system)
 
    - [Build](#build)
+   - [Device type](#devicetype)
    - [Manufacturer](#manufacturer)
    - [Model name](#modelname)
    - [Platform](#platformsys)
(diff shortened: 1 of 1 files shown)


docs/application/native/guides/device/system.md
```diff
@@ -730,6 +730,15 @@ The following table lists the build system keys:
 | `http://tizen.org/system/build.variant`  | `String` | The platform returns the variant release information. The variant release information is made when the platform image is created. |
 | `http://tizen.org/system/build.release` | `String` | The platform returns the build version information. The build version information is made when the platform image is created. |
 
+The following table lists the device keys:
```
mijin85cho: Question. Don't we need to add a link to this table on line 73?

kichankwon: I just add a link like other keys.
If you have a better idea (where this key include), please tell me.

Meanwhile, table name is modified (device -> device type) because it is not device.type but device_type.

mijin85cho: I can see `<a name="devicetype"></a>` is applied well, but for now "#devicetype" is called from nowhere.
So I suggested to add a link to the list in the above of this page like other tables.
image 

The md file, line 73.
image

I think the the location of this table is OK.

kichankwon: Modified.

PR #726, opened 2019-05-28
change supported OS version

winstone77:
Change Supported OS Version

Signed-off-by: seok.oh <seok.oh@samsung.com>


mijin85cho:
Question. From when this update should be published on TD? From right now?

PR #843, opened 2019-09-27
Change the order of menu tree in terms of priority.

journey2w:
### Change Description ###

Change menu tree based on priority.
- Wearable, TV, and Mobile
- .Net App, Web App, Native App


journey2w:
toc_all conflict was resolved.

PR #969, opened 2019-12-06
Change the order of creating applications in overview

journey2w:
### Change Description ###

1. Change the order of creating applications in overview.
2. Fix typo

### DIFF
--- a/docs/application/vstools/index.md
+++ b/docs/application/vstools/index.md
@@ -40,7 +40,7 @@ Visual Studio Tools for Tizen provides Tizen-specific tools to improve your prod
 
 - [Memory Profiler](tools/memory-profiler-user-manual.md)
 
-  This is a utility which collects memory profiling data and save them to a file, this part of the Memory Profiler runs on the target side. Also it analyzes the resulting data, the command-line one and the GUI-based one, these analyzing tools run on the host side.
+  This is a utility which collects memory profiling data and saves them to a file, this part of the Memory Profiler runs on the target side. Also, it analyzes the resulting data, the command-line one and the GUI-based one, these analyzing tools run on the host side.
 
 If you prefer a command line interface to the graphical Visual Studio window, you can use the [Command Line Interface (CLI) tool](tools/dotnet-cli-ext.md) during the entire development process.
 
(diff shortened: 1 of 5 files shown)


docs/application/vstools/index.md
```diff
@@ -40,7 +40,7 @@ Visual Studio Tools for Tizen provides Tizen-specific tools to improve your prod
 
 - [Memory Profiler](tools/memory-profiler-user-manual.md)
 
-  This is a utility which collects memory profiling data and save them to a file, this part of the Memory Profiler runs on the target side. Also it analyzes the resulting data, the command-line one and the GUI-based one, these analyzing tools run on the host side.
+  This is a utility which collects memory profiling data and saves them to a file, this part of the Memory Profiler runs on the target side. Also, it analyzes the resulting data, the command-line one and the GUI-based one, these analyzing tools run on the host side.
```
mijin85cho: > "This is a utility which collects memory profiling data and saves them to a file, this part of the Memory Profiler runs on the target side."

**Suggestion**. Isn't it better to cut this sentence into two in the middle?
This is a utility which collects memory profiling data and saves them to a file. This part of the Memory Profiler runs on the target side.

> "Also, it analyzes the resulting data, the command-line one and the GUI-based one, these analyzing tools run on the host side."

**Suggestion**. How about changing this sentence like below?
 Also, it analyzes the resulting data, the command-line one and the GUI-based one, which run on the host side.

docs/application/web/get-started/overview.md
```diff
@@ -1,36 +1,30 @@
...
-    shown by specific UI applications (such as the home screen).
+    A wearable widget application is a small application that can be shown by specific UI applications (such as the home screen).
 
 -   [TV Web application](tv/first-app.md)
 
     A TV application is targeted to run on smart TVs, and can be optimized for larger screens and remote control.
 
-The first application examples show a simple application with a basic UI
-and minimal functionality. Get familiar with the example first, and
+-    [Samsung Smart TV Web application](first-samsung-tv-app.md)
```
mijin85cho: This link is broken.

I think it should be \(tv/first-samsung-tv-app.md)

docs/application/vstools/index.md
journey2w: Your suggestions are good but the current paragraph is not bad for the short summary.

PR #990, opened 2020-01-02
[WebAPI] Removed unneeded files

pkosko:
Two files are not needed, they are *.orig files created by git during merge conflict
in the past. They should not be added to repository.


mijin85cho:
Thanks for the update @pkosko 
Would you please get approval from one of your team member?

PR #1077, opened 2020-04-09
[6.0][WebAPI] Deprecation of Bookmark module

pkosko:
Whole module deprecation for mobile profile.

[ACR] https://code.sec.samsung.net/jira/browse/TWDAPI-257

* updated bookmark.html
* updated index.html
* updated index.xml

### DIFF
--- a/docs/application/web/guides/personal/bookmarks.md
+++ b/docs/application/web/guides/personal/bookmarks.md
@@ -1,7 +1,6 @@
 # Bookmarks
 
-> **Note**
->
+> [!NOTE]
 > Since Tizen 6.0, the Bookmark API is deprecated and using it is not recommended.
 
 You can manage the Tizen Web browser bookmarks by retrieving the current bookmark list. You can create or delete both bookmark folders and items.
--- a/docs/application/web/guides/personal/bookmarks.md
+++ b/docs/application/web/guides/personal/bookmarks.md
@@ -1,6 +1,6 @@
 # Bookmarks
 
-> [!NOTE]
+> [!CAUTION]
 > Since Tizen 6.0, the Bookmark API is deprecated and using it is not recommended.
 
 You can manage the Tizen Web browser bookmarks by retrieving the current bookmark list. You can create or delete both bookmark folders and items.
--- a/docs/application/web/guides/personal/bookmarks.md
+++ b/docs/application/web/guides/personal/bookmarks.md
@@ -1,5 +1,9 @@
 # Bookmarks
 
+> **Note**
+>
+> Since Tizen 6.0, the Bookmark API is deprecated and using it is not recommended.
+
 You can manage the Tizen Web browser bookmarks by retrieving the current bookmark list. You can create or delete both bookmark folders and items.
 
 This feature is supported in mobile applications only.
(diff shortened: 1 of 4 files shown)


mijin85cho:
Question. We have web guide page also for bookmark.

https://github.com/Samsung/tizen-docs/blob/tizen_6.0_prepare/docs/application/web/guides/personal/bookmarks.md

Do you want to keep this page even the Bookmark is deprecated on Tizen 6.0? I think this page should be deleted also

pkosko:
@mijin85cho, currently module is only marked as deprecated. So basically the guide can be useful in case of maintaining some existing applications. We can remove a guide if you request it, I just would like to confirm it. API is marked as deprecated, but documentation still exists - should we remove a guide?

mijin85cho:
@pkosko As Bookmark is marked as deprecated, not deleted yet, 
how about just adding a note to the starting point of bookmark guide page that bookmark is deprecated since Tizen 6.0?

As far as I know, the Bookmark api will be removed after 2 version (since the deprecation.). I think the guide page can be removed at that time together with the API page.

If you have other opinion, please let me know.

pkosko:
@mijin85cho I added note to guide of Bookmark API.

docs/application/web/guides/personal/bookmarks.md
```diff
@@ -1,5 +1,9 @@
 # Bookmarks
 
+> **Note**
+>
+> Since Tizen 6.0, the Bookmark API is deprecated and using it is not recommended.
```
mijin85cho, line 5: Thanks for the update.

However, the note style we use in Tizen docs is changed.
Would you please check https://github.com/Samsung/tizen-docs/blob/master/styleguide/custom-style.md and change the note style?

\> [!NOTE]
\>Since Tizen 6.0, the Bookmark API is deprecated and using it is not recommended.

Sorry for the late sharing this.

pkosko: I updated this, but it looks like it is not generated in the same way as presented in https://github.com/Samsung/tizen-docs/blob/master/styleguide/custom-style.md
in file:
https://github.com/Samsung/tizen-docs/blob/6af9bd943ccf001cd691cc7a39977307f291035d/docs/application/web/guides/personal/bookmarks.md

pkosko: Even though, it looks good when checking on https://docs1.stg.tizen.org/staging/1077/application/web/guides/personal/bookmarks/

rwalczyna: We don't want people to use it anymore - that's why we are deprecating it, so maybe "Caution" would be more suitable - red color attracts attention more than blue.
What do you think about that?

mijin85cho: @pkosko As you said "View file" doesn't convert well this new style yet.
But as you also checked, it will be converted well on docs.tizen.org. So Don't need to worry about that.

@rwalczyna I think your suggestion to use "Caution" is good. "Caution" can be used instead of "Note".

pkosko: @mijin85cho @rwalczyna Done

PR #1103, opened 2020-05-11
Update IoT PIO Guides from RPi 3 to RPi 3/4

kkuchciak:
This patch updates Peripheral-IO guides from Raspberry Pi 3 to version 4.

Signed-off-by: Konrad Kuchciak <k.kuchciak@samsung.com>


kkuchciak:
Hello, @mijin85cho 

> Are you planning to update these pages also?

iot/guides/iot-api.md refers Get Started pages which for now are only available for RPi3.
I will update iot-api.md when Mr. Jin Yoon create Get Started pages for RPi4.
 
> After this update, peri_api_raspberry.png seems not used from anywhere. If so, how about deleting this not-used image?

Deleted.

mijin85cho:
@kkuchciak Thanks for the checking.

For now, the strings are changed to "Raspberry Pi 3/4", and it seems like to show that both Raspberry Pi 3 and Raspberry Pi 4 have same values for the corresponding context.

If my understanding is correct, how about using "or" instead of "/"?

"Raspberry Pi 3 or 4"

Would you please let me know your opinion?

kkuchciak:
Hi,
Yes, the content of these pages apply both to Raspberry Pi 3 and 4.
Personally, I would prefer to use "Raspberry Pi 3/4" version as this is what has been agreed with my HQ PL.

mijin85cho:
Hello, @surbhig12 

Raspberry 4 information is added to the existing content.

The existing expression "Raspberry Pi 3" is changed to "Raspberry Pi 3/4",
to indicate that the content is both valid for Raspberry Pi 3 and Raspberry Pi 4.

Would you please check the "Raspberry Pi 3/4" expression and let us know your opinion?

If you accept the expression please approve this PR. I'll merge it.

If you have other opinion, please leave a comment. In that case, I hope @kkuchciak also checks again. I appreciate @kkuchciak's answer, but personally not sure with "Raspberry Pi 3/4" so hope to get @surbhig12's checking also.

surbhig12:
Hello @mijin85cho,

I agree with you and in my opinion too Raspberry Pi 3/4 is not the right way to mention. We must elaborate and specify as Raspberry Pi 3 and Raspberry Pi 4
Another reason to write this is also because both versions have sub models, such as Raspberry Pi Model B. So, to be more clear, I suggest we elaborate.

kkuchciak:
@surbhig12 @mijin85cho 
Thanks for your opinion. Let's do as you suggested.

mijin85cho:
Thanks for applying our opinion, @kkuchciak.

@surbhig12 As we asked your opinion also, would you please approve this PR?
For these section I think checking only the updated string would be enough.

PR #1105, opened 2020-05-12
Add missing description about ui_service.sticker feature

Inhong:
### Change Description ###
Add missing description about ui_service.sticker feature

### DIFF
--- a/docs/application/native/guides/device/system.md
+++ b/docs/application/native/guides/device/system.md
@@ -59,6 +59,7 @@ To obtain the information, query a feature or system key.
    - [Shell](#shell)
    - [Sip](#sip)
    - [Speech](#speech)
+   - [Sticker](#sticker)
    - [Storage](#storage)
    - [System setting](#systemsetting)
    - [Thermistor](#thermistor)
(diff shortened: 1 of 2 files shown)


docs/application/native/guides/device/system.md
```diff
@@ -658,6 +659,15 @@ The following table lists the speech feature keys:
...
 | `http://tizen.org/feature/speech.synthesis` | `bool` | The platform returns `true` for this key, if the device supports speech synthesis (TTS). |
 
+The following table lists the sticker feature keys:
+
+<a name="sticker"></a>
+**Table: Sticker feature keys**
+
+| Key                                           | Type   | Description                              |
+|-----------------------------------------------|--------|------------------------------------------|
+| `http://tizen.org/feature/ui_service.sticker` | `bool` | The platform returns `true` for this key, if the device supports Sticker API. |
```
mijin85cho, line 669: I have a question about the order of this API.

The feature is named "Sticker feature", so it is located after "Speech".

However, the feature itself is "ui_service.sticker".
According to the feature itself, it should be considered as "UI service feature" and located before "USB".

Is there no more plan to add kind of "UI service feature"?
I'm not sure that the sticker would be the only feature for "ui_service".

mijin85cho:
Hello, @Inhong 
Would you please check my comment and reply to it?

docs/application/native/guides/device/system.md
Inhong: I tried to support the name sticker, but the reviewer said it would be better to change it to ui_service.sticker, so the name was changed to ui_service.sticker. so I'm not sure if other services will be added to ui_service.
Do I need to move the location of the "Sticker featue"?

mijin85cho: As the feature is ui_service.sticker, in my opinion moving the location is better.
The feature keys section is following the alphabetical order of the feature keys.

My suggestion is to change as below and move to before "USB".

---

...
\- \[Thermistor](#thermistor)
**\- \[UI Sticker](#sticker)**            _(moved location)_
\- \[USB](#usb)
...

---

...
_Thermistor table_

The following table lists the sticker feature keys:

\<a name="sticker"></a>
\*\*Table: UI sticker feature keys\*\*

\| Key                                           | Type   | Description                              |
\|-----------------------------------------------|--------|------------------------------------------|
\| \`http://tizen.org/feature/ui_service.sticker` | \`bool\` | The platform returns \`true\` for this key, if the device supports Sticker API. |

_USB table_
...

---

Would you please let me know your opinion?

PR #708, opened 2019-05-14
Updating Native Get Started Overview

Reni-90:
### Change Description ###

Following are the changes:

- Restructured the content for the overview.md file.
- Added images.

PR #1130, opened 2020-06-02
Updating Native Overview file

Reni-90:
### Change Description ###

Updated content for the overview.md file.

For images and design reference, see #708

### DIFF
--- a/docs/application/native/get-started/overview.md
+++ b/docs/application/native/get-started/overview.md
@@ -1,22 +1,24 @@
 # Get Started with Native Applications
 
-When creating your first Tizen native application, you can select from the following options, which all use the C language:
+Tizen Native applications are developed in C. You can use the available native API modules to have advanced system settings and additional functionalities.
 
--   [Wearable native application](wearable/first-app.md)
+To create your first Tizen Native application, the following profiles are available:
 
-    A wearable application is targeted to run on smart watches, and can be optimized for the smaller screen.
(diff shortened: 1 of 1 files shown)


docs/application/native/get-started/overview.md
```diff
@@ -1,22 +1,24 @@
 # Get Started with Native Applications
 
-When creating your first Tizen native application, you can select from the following options, which all use the C language:
+Tizen Native applications are developed in C. You can use the available native API modules to have advanced system settings and additional functionalities.
```
journey2w, line 3: I understood what you said. I guess you'd like to describe Tizen Native application first, but your explanation is a little awkward.

I'd like to suggest as follows:
```suggestion
You can create your Tizen application  with Tizen native APIs, which are written in C language and provide Tizen additional functionalities.
```

If you want to add the definition of Tizen native application, please don't hesitate.

Reni-90: Thank you for understanding. I have taken your suggestion and tweaked a little.
Can we rephrase as:

journey2w: Can I add something here?
The first sentence is an introduction and a key sentence.  
Some suggested sentences are the first sentence on the linked pages starting from "Create...".
We need a change. 

Are there better sentences that introduce Native, Web, .NET applications respectively?
What a Tizen XXX App is and developers can do what with it.

PR #1146, opened 2020-06-15
Add menu.yaml

huytaduc:
### Change Description ###

GNB settings by YAML file

### Bugs Fixed ###

N/A

### API Changes ###

N/A

### DIFF
--- a/docs/menu.yaml
+++ b/docs/menu.yaml
@@ -1,0 +1,16 @@
+- name: "Platform"
+  href: "/platform/what-is-tizen/overview"
+
+- name: "Application"
+  href: "/application"
+
+- name: "IoT"
+  href: "/iot"
+
+- name: "Partners"
+  href: "/partners/specialist"
+
+- name: "Download"
+  href: "https://developer.tizen.org/development/tizen-studio/download"
+  external: true
+  


docs/menu.yaml
```diff
@@ -0,0 +1,16 @@
...
+- name: "IoT"
+  href: "/iot"
+
+- name: "Partners"
+  href: "/partners/specialist"
+
+- name: "Download"
+  href: "https://developer.tizen.org/development/tizen-studio/download"
+  external: true
+
```
mijin85cho, line 16: **Question:**

It seems we can create & edit the GNB menu and control which page will be displayed under each menu.
Is the each "href" showing the path to the "index.md" page that appears first when the each menu are clicked?

Also in case that we don't have "index.md" file, 
it seems that we can designate other file instead.

ex)   href: "/platform/what-is-tizen/overview"

In this case the file name extension, `.md` is excluded. Is my understanding right?

huytaduc: @mijin85cho 
In the case we don't have `index.md` we just need to remove /index from the path.
Below are the rules to convert from file path to URL:

- remove .md from filename
- if file name is index.md remove `index.md`

For example:
```
- platform
       - what-is-tizen
             - index.md        => /platform/what-is-tizen
             - overview.md    => /platform/what-is-tizen/overview
             - test.md           => /platform/what-is-tizen/test
- application
        - index.md             => /application
        - test.md                => /application/test
        ....
```

PR #1157, opened 2020-06-25
Tizen 6.0 prepare

haejeongkim:
Since 6.0, metadata extractor supports WAV, FLAC and OGG files as well.
https://code.sec.samsung.net/jira/browse/ACR-1569(but No API Changes.)


docs/application/native/guides/multimedia/metadata.md
```diff
@@ -7,7 +7,7 @@ To handle content metadata, you can use the following features:
 
   You can [edit the metadata](#edit_metadata) of several popular audio formats using the metadata editor. You can add and remove album art, or update the information for the audio file.
 
-  The metadata editor supports editing the [metadata](#attribute) of the mp3 and mp4 (audio only) file formats. Image and video file editing is not supported.
+  The metadata editor supports editing the [metadata](#attribute) of the MP3, mp4, WAV, FLAC and OGG file formats. Image file editing is not supported.
```
mijin85cho, line 10: Question. 

- Don't we need to change mp4 to MP4 as like the mp3 case?
- Also for OGG, when I searched on Google it seems to write as "Ogg". Would you please check?
- We have "mp3 and mp4" on line 14. For the consistency how about changing them as "MP3 and MP4"?

Please add a comma after FLAC.
...WAV, FLAC, and OGG file formats.

PR #1185, opened 2020-07-23
[fix] Remove items (and it's descriptions) deprecated 2+ versions ear…

djuszczakk:
Some items are deprecated since version 5.0 or earlier but were still in the documentation.
This commit removes these items and its descriptions


mijin85cho:
Thanks for the update, @djuszczakk. For Web api files, to proceeding to the next step, would you please get cross check from your team member?

PR #1186, opened 2020-07-23
[fix] Remove items (and it's descriptions) deprecated 2+ versions earlier

djuszczakk:
Some items are deprecated 2 or earlier version since generated html
but are still in the documentation.
This commit removes these items and its descriptions.
Also fixes some minor issues


mijin85cho:
Thanks for the update, @djuszczakk. For Web api files, to proceeding to the next step, would you please get cross check from your team member?

PR #1215, opened 2020-08-26
Fix update control docs

sh95jeon:
### Change Description ###

- Current update_get_property usage example is wrong (causing segmentation fault), so fixed it.

### Bugs Fixed ###

- No related bug

### API Changes ###

- Non ACR


mijin85cho (approved):
Thanks for the update.

Issue #1222, opened 2020-09-01
Need to update SDK prerequisite guide


poiema72:
https://docs.tizen.org/application/tizen-studio/setup/prerequisites/

From above, current Tizen Studio supports only Windows 10.

However, T-trace guide includes window 7/8 like below.

`To use the T-trace in Windows® 8/7, you must install a Python 2.7.X version:
On the Python Web site, download the appropriate Python version for your hardware and Windows® version.
Run the downloaded installer file and follow the displayed instructions.`

I think that it needs to be revised.

mijin85cho:
Thanks for the issue report.

Related with this, https://github.com/Samsung/tizen-docs/pull/1223 is merged.

We will merge the update to live branch later, and it will update docs.tizen.org.

mijin85cho:
As the fix for this PR is merged, I close this PR. Thanks for the report.

PR #1223, opened 2020-09-01
fix wrong version info

mijin85cho:
Signed-off-by: Mijin, Cho <mijin85.cho@samsung.com>

### Change Description ###

Update the wrong working window version in Tizen Studio prerequisites page.

### Bugs Fixed ###

https://github.com/Samsung/tizen-docs/issues/1222

### DIFF
--- a/docs/application/tizen-studio/setup/prerequisites.md
+++ b/docs/application/tizen-studio/setup/prerequisites.md
@@ -186,7 +186,7 @@ The following table lists the additional requirements to be met before developin
 <tr>
 <td>Python for using the T-trace (Tizen profiling tool used to optimize the application performance)
 </td>
-<td>To use the T-trace in Windows&reg; 8/7, you must install a Python 2.7.X version:
+<td>To use the T-trace in Windows, you must install a Python 2.7.X version:
 <ul>
 <li>On the <a href="https://www.python.org/downloads/">Python Web site</a>, download the appropriate Python version for your hardware and Windows&reg; version.
 </li>


docs/application/tizen-studio/setup/prerequisites.md
```diff
@@ -186,7 +186,7 @@ The following table lists the additional requirements to be met before developin
 <tr>
 <td>Python for using the T-trace (Tizen profiling tool used to optimize the application performance)
 </td>
-<td>To use the T-trace in Windows&reg; 8/7, you must install a Python 2.7.X version:
+<td>To use the T-trace in Windows, you must install a Python 2.7.X version:
```
annie-samsung, line 189: @mijin85cho spoke to Karthik, he suggested to mention like this:

To use the T-trace in Windows&reg; 10 and above, you must install a Python 2.7.X version:

mijin85cho: Question.

Isn't the "least" version for the Tizen Studio itself the Windows 10?
Thus there is no other choice except "Windows 10 and above" for now, so I didn't mention in previous update.

image

mijin85cho: @annie-samsung Would you please let me know you opinion to my comment?

annie-samsung: @mijin85cho the idea here was to keep the document future proof, so that we don't have to update going forward for next Windows version. However, if you want we can just mention Windows 10 for now, and remember to change it when another Windows version is also supported.

mijin85cho: @annie-samsung It is a good idea to prepare the future version, but I think it is better to change in that way if the Tizen Studio "really" supports the future version. For now, "the future version" doesn't exist, and I can't sure that Tizen Studio will support that version from when. (Maybe it supports immediately when future version is released. But I think it also has the possibility that further working is needed to make Tizen Studio compatible with the "above Windows 10". I'm not sure with it.)

annie-samsung (approved):
Approving the document, after review.

mijin85cho:
Thanks for the checking, @annie-samsung

PR #1224, opened 2020-09-01
Add new privilege: network.route

yl33:
### Change Description ###

- New privilege http://tizen.org/privilege/network.route was added for
Native and dotnet application by following ACRs:
  - https://code.sec.samsung.net/jira/browse/ACR-1586
  - https://code.sec.samsung.net/jira/browse/TCSACR-360


mijin85cho:
Just a question.

| `http://tizen.org/privilege/network.route` | partner |  | The application can add or remove route table entries. |

Isn't "routing table" more common term than "route table"?

yl33:
> Just a question.
> 
> | `http://tizen.org/privilege/network.route` | partner | | The application can add or remove route table entries. |
> 
> Isn't "routing table" more common term than "route table"?

Both of them are commonly used in technical docs (MS, Amazon, and Oracle also use term "route table" with "routing table") so I think there would be no difference between them.

mijin85cho (approved):
Thanks for the answer @yl33 

As this is not big update, I skip the review from SRI-B.

PR #1245, opened 2020-09-15
Updating images oneUI

surbhig12:
### Change Description ###

As part of the oneUI image updation: Updated the following four images in the `docs\application\vstools\tools\media` folder:

- emulator_tv_mobile.png
- memory_profiling_app_started
- profiler_workspace
- run_profiler.png


mijin85cho:
Checkpoints: 

- In the updated images, the project names differ from the original ones.
   
   |Image|Project name in Original|Project name in Updated|
   |--|--|--|
   |memory_profiling_app_started.png|CrossTemplate1|CrossTemplate3|
   |profiler_workspace.pngg|NUITemplate1|NUITemplate2|
   |run_profiler.png|NUITemplate1|NUITemplate3|

- For example in https://docs1.stg.tizen.org/staging/1245/application/vstools/tools/profiler-user-manual/, 
   - in the original images the workspace is C:\Users\tizenuser\source\repos\NUITemplate1
   - in the updated images the workspace is C:\Users\surbhi.gupta\source\repos\CrossTemplate3

It would be better to unify the workspace information in the images (ex. in the logs..),
and moreover, I strongly recommend you to remove your personal information (surbhi.gupta) from the image.
Using the existing location, C:\Users\tizenuser\source\repos would be better.

surbhig12:
@mijin85cho, thank you for your suggestions. I have updated the images and also removed personal information.

mijin85cho:
Thanks for the update. However I want to ask your opinion about profiler_workspace.png.

In the image, there is a Profiling Progress window with red highlight.

Comparing to the original image, it seem the Profiling Progress is stopped almost right after(under 5 secs) starting it.

|Original Image | Updated Image |
|-|-|
|image |image|

For better view, how about updating the image with longer processed window?
Would you please let me know your opinion related with this?

surbhig12:
@mijin85cho, based on the sample app that we have, this was the output we got (at 5 seconds). We could use this image for now and later when major changes happen we can replace the image.

mijin85cho:
> @mijin85cho, based on the sample app that we have, this was the output we got (at 5 seconds). We could use this image for now and later when major changes happen we can replace the image.

I mean, as like the original one, how about using the image you got at about 60 seconds? For now the result graph is too short comparing to the original one.

surbhig12:
@mijin85cho, I wasn't able to get any image for 60 seconds. Please suggest how shall we proceed on this?

mijin85cho (approved):
Thanks for the updates.

PR #1252, opened 2020-09-17
Updating images oneUI

surbhig12:
### Change Description ###

As part of the oneUI image updation, updated the emulator_running_tv image in the `docs\application\web\get-started\tv\media` folder.


mijin85cho:
Hello, @surbhig12 
I have a question to this update.

In https://docs1.stg.tizen.org/staging/1252/application/web/get-started/tv/first-app/#running-on-the-emulator,
I can see there is an emulator image which is using the old emulator version, that shown on emulator_running_tv.png, which is replaced by you.

**Image name :** emulator_window_tv.png

Don't we need to check this image also?

surbhig12:
Hello @mijin85cho,

As of now, we were only focusing on the images that were listed in the excel sheet shared by @journey2w. If we need to update other images then the list would be very long. Please let us know how to proceed further. Shall we update all the imges in the files or stick to the images listed in the excel?

mijin85cho:
Hello, @surbhig12 
Indeed, checking every image and updating them is ideal - and impossible realistically.

So I think focusing on the excel file is the way we should choose.

In case of emulator_window_tv.png, I asked to check it particularly, because it has the emulator image - and the emulator image is outdated.

As emulator_running_tv.png was asked to be updated due to the same reason, the old emulator image,
I thought it would be a good chance to update emulator_window_tv.png also.

surbhig12:
@mijin85cho,

Thank you for your response. I will look into this file and change the image.

mijin85cho:
@surbhig12 Thanks for your reply. Just to prevent the confusion - With my previous comment, I'm asking only to check  emulator_window_tv.png, not other all images in the folder : )

surbhig12:
@mijin85cho, yes, I will check the emulator_window_tv.png alone :)

surbhig12:
@mijin85cho, the emulator_window_tv.png image can currently not be updated as it is device dependant. Therefore, we will stick to the excel for now.

mijin85cho (approved):
Thanks for the update.

PR #1254, opened 2020-09-17
Updating images oneUI

surbhig12:
### Change Description ###

As part of the oneUI image updation, updated the following images in the `docs\application\vstools\media` folder:

- cps-extensions-and-updates
- cps-project-property


mijin85cho (approved):
Thanks for the update.

PR #1257, opened 2020-09-21
Updating images oneUI

surbhig12:
### Change Description ###

As part of the oneUI image updation, updated the 1.1.1.png image in the `docs\application\design\mobile\media` folder.


mijin85cho:
Just a question. 
In original image it seem the performance is adding the more menu from the menu button, to the "current screen".

The sentence in the context : 
_...If you need to add subordinate actions in the design, provide them in ‘More’ menu...._

However in the updated image,
it seems the calendar image in the background is different between before touching the "menu" and after the touching.
Does it mean if you touch the "menu" button, the calendar UI is changed?

surbhig12:
@mijin85cho, we need a sample device to test this. I suggest that we close this PR for now and mark this image as unable to reproduce in the excel.

mijin85cho (approved):
Thanks for the update.

PR #1259, opened 2020-09-22
Updating images oneUI

surbhig12:
### Change Description ###

As part of the oneUi image updation, updated the cs_first_creating_project_wizard_tv.png in the `docs\application\dotnet\get-started\tv\media` folder.


mijin85cho:
Thanks for the updates, @surbhig12.

I have one doubt. I agree with that your update is more proper according to the context.

However in the excel file, it seems there is a comment for this image at the column C : "To change Wearable (preview)" 
It sounds like they are asking to update this image, because of the "Wearable (preview)".

I'm not sure with Visual Studio 2019, that (preview) is removable. Could you please check in this point once again?

mijin85cho (approved):
Thanks for the update.

PR #1288, opened 2020-10-14
Update event.md

mokind:
bundle_get_val() is not available in bundle.h (anymore?), but in this example it is a string value, so the change should do it.

### Change Description ###

Describe your changes here.

For instance,
 -  ...

### Bugs Fixed ###

Provide links to bugs here

For instance,

 - Issue #265

### API Changes ###

For instance,

 - ACR: ACR-1120


mijin85cho (approved):
Thanks for the update.

Issue #1289, opened 2020-10-16
Native API reference UI issue


swift-kim:
If you search for some Tizen API on Google such as "system_info_get_platform_int", normally Tizen 2.3.1 mobile API references come at the top.

https://docs.tizen.org/application/native/api/mobile/2.3.1/group__CAPI__SYSTEM__SYSTEM__INFO__MODULE.html

As you can see in the above page however, there's no side menu in the left side of the page. As a result I can't switch to a different profile or newer API version.

![image](https://user-images.githubusercontent.com/40190588/96211476-66018080-0faf-11eb-91d3-e8afa417f7b2.png)

It looks normal if I change "2.3.1" to "5.5" in the URL.

https://docs.tizen.org/application/native/api/mobile/5.5/group__CAPI__SYSTEM__SYSTEM__INFO__MODULE.html

![image](https://user-images.githubusercontent.com/40190588/96211555-977a4c00-0faf-11eb-92d5-2b93cddfa03f.png)

- Could you just redirect to the latest (5.5) API if 2.3.1 is officially unsupported?
- Could you add the side menu in the old API references page?

mijin85cho:
I reported this issue with https://github.sec.samsung.net/RS8-TIZENWEBSITE/docs.tizen.org/issues/413.

journey2w:
We won't show 2.3.1 version on the API reference page.
Resolved on PR #1325.

PR #1325, opened 2020-10-30
Remove Tizen 2.3.1 Native Reference for mobile and wearable

journey2w:
### Bugs Fixed ###

- Problem
  I can't see LNB of https://docs.tizen.org/application/native/api/wearable/2.3.1/index.html and https://docs.tizen.org/application/native/api/mobile/2.3.1/index.html because we don't provide a menu item for 2.3.1 version

- Resolution
  I removed files to prevent unexpected access.
  We have archive files for then in archive directory.
  Anytime we want to show this version, we can do it.

  This PR is resolved issue #1289


mijin85cho (approved):
Thanks for the update.

PR #1369, opened 2020-12-09
Deprecate Activity & Gesture Recognition

pvschakradhar:
Deprecating Activity & Gesture Recognition files because capi-context-motion is deprecated
ACR: ACR-1588 (https://code.sec.samsung.net/jira/browse/ACR-1588)


annie-samsung:
@ pvschakradhar, I understand that this PR is about removing two files: Activity and Gesture Recognition
What about other documents (if any) linking to these documents? Those documents need to be identified and links need to be removed. 

@mijin85cho, thoughts?

mijin85cho:
> @ pvschakradhar, I understand that this PR is about removing two files: Activity and Gesture Recognition
> What about other documents (if any) linking to these documents? Those documents need to be identified and links need to be removed.
> 
> @mijin85cho, thoughts?

To remove a page, we should check the links also for the case that other page has a hyperlink to the removed page.

For example, the application/native/guides/index.md and application/native/guides/location-sensors/overview.md pages have links to removed pages, 
so these links should be removed also. Otherwise we'll get broken links.

annie-samsung:
@pvschakradhar could you please check on the above points

mijin85cho:
Thanks for updating.

However, if there were inserted images in the deleted page and the images are not used in other pages, would you please remove the images also?
For example in gesture.md, there are 7 images. As these images are not is use now, they should be removed also.

@annie-samsung I hope you consider this point also.

annie-samsung (approved):
Checked for dependencies, no other files seem to be impacted. Hence approving.

PR #1409, opened 2021-03-02
Update iot tcs 0.98 beta

sgchoi5:
Replaced iot tcs 0.97 beta with 0.98 beta


annie-samsung (approved):
Looks good, hence approving

PR #1420, opened 2021-03-25
Fix wrong image and description

KwangSon:
### Change Description ###
VS for mac and VS code image and description are exchanged.

Signed-off-by: Kwang Son <k.son@samsung.com>


annie-samsung (approved):
looks good hence approving

PR #1424, opened 2021-04-08
[NSD] Fix wrong API names in Network Service Discovery Guides

chleun-moon:
### Change Description ###

Fix wrong API names and errors.


annie-samsung (approved):
looks good, hence approving

PR #1449, opened 2021-05-25
Add new removed functions to list

sgchoi5:
All video util functions are removed.


annie-samsung (approved):
looks good hence approving

PR #1458, opened 2021-06-04
Update localization.md

nawab-samsung:
test
KT


nawab-samsung:
This was just for testing

PR #1496, opened 2021-08-11
[ML] Renamed a SRNPU with TRIX_ENGINE

pkosko:
[ACR] https://code.sec.samsung.net/jira/browse/TWDAPI-280


annie-samsung (approved):
Looks good

PR #1514, opened 2021-09-14
Fixing headers format issue in topics (docs/platform/developing)

sagarkedari:
Since headers in these topics/files were written in title case and gerund forms, I have updated the headers as per Tizen writing style guidelines.


annie-samsung (approved):
look good

PR #1568, opened 2021-12-08
Revert "Whatistizen65"

annie-samsung:
### Change Description ###

Describe your changes here.

For instance,
 - Fix broken links of native application docs.
 -  ...

### Bugs Fixed ###

Provide links to bugs here

For instance,

 - Issue #265

### API Changes ###

In case of ACR, put the link of ACR

For instance,

 - ACR: ACR-1120

PR #1571, opened 2021-12-14
Revert "Revert "Whatistizen65""

annie-samsung:
Reverts #1568

Reverting 1568 to avoid the confusion


annie-samsung:
@mijin85cho, hope I am doing this correctly, I need to merge this PR now to undo the deletion I did of 1554, is that correct?

mijin85cho:
> @mijin85cho, hope I am doing this correctly, I need to merge this PR now to undo the deletion I did of 1554, is that correct?

Hello, @annie-samsung, yes you're right. You can also see in the file diff that removed pages are added back with this PR 1571.

mijin85cho (approved):
To revert removed WhatisTizen6.5

PR #1650, opened 2022-06-22
Fixed privilege for Content Downloads

DavyDeCoster:
### Change Description ###

Describe your changes here.

- Deleted the extra space in Content Downloads guide which give the privilege to add to the config.xml. Just a simple copy-paste would break the build.


mijin85cho:
Dear SRBD reviewer

As you'd see, this update is simple - just removing an unnecessary space.
It is small, but important fix, so you can focus only to the update.

**FYI :** In case of source code, 
the document guide must be accurate so that the developers can follow the guide just by copying & pasting the codes.

safir-srbd (approved):
@DavyDeCoster Thank you for the update.

@mijin85cho The change made by the author seems to be reflected correctly on the staging site, so I am approving this PR for merging.

PR #1698, opened 2022-09-30
.NET application update (Overview, Learn about .NET & Get Started)

omar-srbd:
### Change Description ###

This is the second PR in the process of updating the .NET applications section of Tizen Docs site.

Signed off by: [omar.saeed@samsung.com](mailto:omar.saeed@samsung.com)

PR #1728, opened 2022-11-16
Update first-app.md

jiwonpark:
Fix minor grammatical error


omar-srbd (approved):
@jiwonpark, thanks for the change.

As this file is part of our Tizen docs site fix, all the changes in this file will be reflected through #1698.
I have only checked the changed part and found everything in place.

@mijin85cho @asmfoysal I am approving this PR for merging.

PR #1790, opened 2023-03-16
Fix default for crash manager crash reports root path

arkq:
### Change Description ###

It seems that `/opt/user/share/crash/` contains a typo. The default path seems to be `/opt/usr/share/crash/`. This PR fixes that typo.


safir-srbd (approved):
@arkq Thank you for pointing out the typo. Your change is valid here. I am approving this PR for merging.

PR #1943, opened 2024-03-08
[9.0][content][Deprecation] Deprecate members not used frequently

pkosko:
Native members were deprecated, thus deprecate also on web side.

[ACR] https://jira.sec.samsung.net/browse/TWDAPI-294


safir-srbd (approved):
@pkosko LGTM. I am approving and merging this PR.

PR #1945, opened 2024-03-20
[Acr-1838] Deprecate all bookmark and playlist funtions

MinjeAhn:
### API Changes ###
 - ACR: [ACR-1838](https://jira.sec.samsung.net/browse/ACR-1838)


safir-srbd (approved):
@MinjeAhn LGTM

PR #1994, opened 2024-07-25
update platform.core.abi information

dkson95:
### Change Description ###

Describe your changes here.

For instance,
 -  ...

### Bugs Fixed ###

Provide links to bugs here

For instance,

 - Issue #265

### API Changes ###

For instance,

 - ACR: ACR-1120


msjang0 (commented):
It looks there are so many duplicates of the same phrase, in the added sentenses. How about simplifying them such as "Thus, "arm_32" will be returned for "arm 32-bit" devices.  Likewise, each of the values "arm_64", "riscv_32",  "riscv_64", "x86", and "x86_64" is used for "arm 64-bit",  "risc-v 32-bit", "risc-v 64-bit", "x86 32-bit", and "x86 64-bit" devices, respectively." ?

safir-srbd (approved):
@dkson95 LGTM!

PR #2010, opened 2024-09-04
Fix example code

jeik01:
The example code in the file below was incorrectly written, so it is corrected.
   - docs/application/native/guides/connectivity/wifi.md


safir-srbd (approved):
LGTM

PR #2012, opened 2024-09-09
Barcode Tutorial Typo Fix

Bhuvan-1:
Obvious typo fix. The variable name should be `image_height`. It is referenced correctly in the subsequent lines.


safir-srbd (approved):
@Bhuvan-1 LGTM.

PR #2019, opened 2024-09-27
[ACR-1862][tts] Add API to support personal voice

psw0405:
[ACR-1862][tts] Add API to support personal voice


safir-srbd (approved):
LGTM

PR #2047, opened 2024-12-10
Remove non-printing characters M-oM-;M-?

kimcinoo:
The `# Title` was broken because of the non-printing characters.
This commit is removing the characters.


maliha-srbd (approved):
PR LGTM

PR #2052, opened 2024-12-19
Update flahsing-rpi.md file

jh80chung:
The location of sd_fusing.py was changed.
Update to new url instead of old url.
- repository was changed from platform/kernel/u-boot to platform/kernel/tizen-fusing-scripts


devusr-sw-kim (approved):
Now, fusing script git repository is crated, so this seems fine.

maliha-srbd (approved):
LGTM

PR #2118, opened 2025-07-24
Remove tizen school from index

jeesunhub:
### Change Description ###
 - tizen schools are removed. contents should be removed.


safir-srbd (approved):
LGTM

PR #2130, opened 2025-08-11
[SystemF/W] Update docs/platform/HAL/system.md

germonado:
### Change Description ###

- Update System Overview
- Update Device Resource Management Section


germonado:
It will be reuploaded.

PR #2140, opened 2025-08-19
Update Web Inspector guide document to the version 139

iRoy7:
### Change Description ###

Update Web Inspector guide document to the version 139


iRoy7:
There are some problems related to uploading image files.

PR #2143, opened 2025-08-19
[Webapi] [10.0] Fixed invalid deprecation version in index.xml file

pkosko:
While preparing index files update for 9.0 version, deprecate version was corrupted - probably with replace-all function.

This change restores original deprecate version == 8.0 for messaging and ppm.

This is the same change as
https://github.com/Samsung/tizen-docs/pull/2142 but for 10.0 API docs structure.


safir-srbd (approved):
LGTM

PR #2146, opened 2025-08-25
Remove the unsupported target flashing guidance

jh80chung:
TM1 is not supporting anymore since Tizen 8.0.
So it doesn't need to maintain this page.
Remove the linking page about flahsing.md.


jh80chung:
I will update again with other commit.

PR #2182, opened 2025-09-22
Update docs/guides/data/data-storages.md to fix incorrect explannation

ul24:
The explanation for 'total size' and 'available size' is written incorrectly and needs to be corrected

### Change Description ###
Update libstorage (capi) and Tizen.System.Storage (csharp) guide to correct what is written incorrectly.

### Bugs Fixed ###
NONE
### API Changes ###
NONE


theojin (approved):
LGTM

PR #2190, opened 2025-09-25
Update docs/guides/data/data-storages.md to fix incorrect explannation

ul24:
The explanation for 'total size' and 'available size' is written incorrectly and needs to be corrected

### Change Description ###
Update data-storages.md to correct the wrong explanation.

For instance,
 -  ...

### Bugs Fixed ###
None

### API Changes ###
None


theojin (approved):
LGTM

PR #2199, opened 2025-09-30
Revert "docs: Update device-sensors.md to add sleep detector"

SangyounKwak:
### Change Description ###

* Since the sleep detector is not in the feature, it should be removed from the document.


theojin (approved):
LGTM

PR #2340, opened 2026-06-12
[Web API] Tizen 10.1 prepare added base

pkosko:
* preapred Tizen 10.1 base for webapi ACRs
  * copy-pasted docs from Tizen 10.0
  * regenerated docs with version update and deprecated API removal
  * updated latest and index file


theojin (approved):
LGTM. Approved.

PR #2351, opened 2026-07-28
Added newly removed modules to history table

sgchoi5:
Added newly removed modules
Updated internal link to ap reference


theojin (approved):
LGTM. Approved.

PR #2359, opened 2026-08-13
docs: fix outdated Tizen Studio setup docs

### LINKED ISSUE #1327: HAXM description needs to be updated.
https://docs.tizen.org/application/tizen-studio/setup/hardware-accelerated-execution-manager/

From the link above, we can find the description below.

_Supported Windows versions: 10/8.1/8/7 (32/64-bit)_

But as of now Tizen Studio only supports windows 10 and 64 bit. Thus considering this, I think that it needs to be updated.

_Ensure that Hyper-V is disabled (on Windows 8 or after)_

The above one seems to be a similar case.

### LINKED ISSUE #1393: install-openjdk-for-windows information missing or wrong
https://github.com/Samsung/tizen-docs/blob/master/docs/application/tizen-studio/setup/openjdk.md#install-openjdk-for-windows
describes how to install open jdk. however it says that you simply have to add JAVA_HOME and after this you can call java from the commandline.
this is wrong. java in openjdk is under the bin directory in the jdk zip. the bin directory is also where the java executable lives. but according to the documentation this path will not be added. you at least have to additionally add the bin directory to the system/user path variable to be able to call java from the commandline.

theojin:
## Summary

Fixes two outdated Tizen Studio setup documents reported in #1327 and #1393.

### #1327 — HAXM supported version lists

`hardware-accelerated-execution-manager.md` listed `Windows 10/8.1/8/7 (32/64-bit)` and macOS 10.8–10.12, both of which are long EOL. Updated to match the current `prerequisites.md`:

- Windows: **10 (64-bit)**
- macOS: **11.4 (Big Sur), 12.3 (Monterey)**

### #1393 — JAVA_HOME path includes \\bin

`openjdk.md` instructed users to set `JAVA_HOME` to `C:\\Users\\user\\Desktop\\jdk-12.0.2\\bin`. `JAVA_HOME` must point to the JDK root directory, **not** the `bin` subdirectory. Changes:

- Removed `\\bin` from the `JAVA_HOME` example and added a note not to include it.
- Added a new step to add `%JAVA_HOME%\\bin` to the **Path** variable so `java`/`javac` are available from the command line (the core of the original report).
- Fixed the macOS `export JAVA_HOME = /Library/...` command — spaces around `=` are invalid in shell assignment.

## Checklist

- [x] Text-only changes; no links added or removed
- [x] Version numbers cross-checked against `prerequisites.md`

Fixes #1327
Fixes #1393

### DIFF
--- a/docs/application/tizen-studio/setup/hardware-accelerated-execution-manager.md
+++ b/docs/application/tizen-studio/setup/hardware-accelerated-execution-manager.md
@@ -15,7 +15,7 @@ Hardware requirements:
 <a name="on_Windows"></a>
 ## Installing Intel&reg; HAXM on Windows
 
-Supported Windows versions: 10/8.1/8/7 (32/64-bit)
+Supported Windows versions: 10 (64-bit)
 
 **HAXM is automatically installed as part of Tizen Studio**. If you want to install the HAXM separately, visit [http://download.tizen.org/sdk/haxm/7.0.0/win/](http://download.tizen.org/sdk/haxm/7.0.0/win/IntelHAXM_7.0.0.exe) and download the Windows installer package.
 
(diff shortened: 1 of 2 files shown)


sangwook96-lee (approved):
오 수고하셨습니다.

kiso-chang (commented):
수고하셨습니다.

PR #2365, opened 2026-08-14
docs: restore 14 pages that are still live on the docs site

theojin:
## What

An SSOT coverage audit compared the published site against `storage/docs` in the publishing pipeline and against this repository. It found pages whose source markdown was deleted here **while the site still serves them with HTTP 200**. On the next full pipeline pass, those pages disappear.

This restores **14 pages**, the **39 media files** deleted alongside them, and the **10 `toc_all.md` entries** they need. The four overview pages are read from fixed paths by their section scripts and build their own `toc.json`, so they take no toc entry.

Every restored URL was verified to return 200 on the live site.

### 1. Swept by the 2026-08-12/13 cleanups (10 pages)

| commit | subject |
|---|---|
| `63609d51d` | docs: remove obsolete unlisted content |
| `edca393fb` | docs: remove Mobile and Wearable profile content |
| `d0e188132` | docs: remove remaining Mobile/Wearable-only tool documentation |
| `25cf65689` | docs: remove RT-IDE, Mobile/Wearable profiles, and resolve all T-ORPHAN orphans |

### 2. Lost to a path mismatch in the IA refactor (4 pages)

`18280fee0` moved these files to `web/guides/concepts/` but registered them in `toc_all.md` as `web/guides/essentials/`:

```diff
-### [Web Runtime](/application/web/tutorials/web-runtime.md)
+### [Web Runtime](/application/web/guides/essentials/web-runtime.md)
```

The files landed in one directory and the TOC pointed at another, so all four were **orphaned the moment they moved**. Four days later `63609d51d`, which removes unlisted content, swept them. They were not deleted on purpose — they were deleted because a rename lost them.

The same mismatch hit four `native/guides/concepts/` files too, but those are 404 on the site, so they are out of scope here.

**Restored at `essentials/`** — the path the refactor declared, not `concepts/` where the files actually went, and not the old `tutorials/` path. `/docs/application/web/guides/essentials/error-handling` is live today and keeps working. The other three change URL, which is what the refactor intended.

## Two side effects of the same refactor, also fixed

- It moved the files one directory deeper without adjusting relative links, so `../../tizen-studio/` resolved to `web/tizen-studio/` instead of `application/tizen-studio/`. **5 links** in `signing-certificates.md` and `web-runtime.md` corrected.
- `web/guides/concepts/media/` held the four images for these pages and was left behind when the markdown was deleted. Moved to `web/guides/essentials/media/`, so no orphaned images remain.

## Deliberately not restored

| file | why |
|---|---|
| `native/guides/applications/ui-app.md` | EFL toolkit content, removed on purpose in 2025-11. `storage/docs` no longer carries the file — only a stale toc node |
| `web/tutorials/old-versioned-sec-privileges.md` | properly relocated to `web/reference/security-privileges-legacy.md`, which this branch restores. Identical content (same blob); the duplicate at the old path is not kept |
| `extensions/tizenx/overview.md` | the tizenx overview section reads `overview/overview.md`, present on master. The deleted file was an older orphan the pipeline never read |

## Relation to `docs/restore-pages-still-live-on-site`

This branch supersedes it. Ten of its eleven pages are included, with its link repairs carried over verbatim. The eleventh is the tizenx file above, which was restored there in error. This branch adds the four IA-refactor casualties, the media files, and the `toc_all.md` entries that branch did not restore.

## Verification

- Every entry added to `toc_all.md` resolves to a restored file
- Every relative link in the restored pages resolves to a tracked file (0 unresolved)
- Restoring markdown alone is not sufficient: 10 of these pages live in sections whose TOC is built from `docs/application/toc_all.md`, and the deletions removed those entries too


theojin:
Correction pushed as `e1c298c4`. The first commit restored three of these pages under `web/guides/essentials/`; they now go back to `web/tutorials/`, where they are actually served.

**Why the first placement was wrong.** I had verified the pages with HTTP 200, which is not sufficient here. The site mirrors `storage/docs` in three states and 200 covers two of them:

| storage | site |
|---|---|
| toc node + markdown | 200 with content |
| toc node, no markdown | **200 with an empty shell** — no page title |
| no toc node | 404 |

Measured on the four pages in question:

```
web/tutorials/event-handling                200, content
web/tutorials/sign-certificate              200, content
web/tutorials/web-runtime                   200, content
web/guides/essentials/event-handling        404
web/guides/essentials/signing-certificates  404
web/guides/essentials/web-runtime           404
web/guides/essentials/error-handling        200, empty shell
```

`guides/essentials/` was never serving those three. Restoring there would have retired three working URLs and published three nobody links to.

`error-handling` stays at `guides/essentials/`. It has no live URL to preserve — the essentials path is an empty shell, and its pre-refactor path `web/guides/error` is 404 — so the location the refactor declared is as good as any, and the toc entry already points there.

The relative-link depth fix from the first commit is reverted for the two files that moved back: `web/tutorials/` is one level higher than `web/guides/essentials/`, so `../../tizen-studio/` is correct again. Media moved with them.

Net effect on the PR: same 14 pages, three at different paths. All local links resolve and every page is listed in `toc_all.md`.

mindol (approved):
Looks good to me.

PR #2366, opened 2026-08-16
docs(toc): put entries in the block that can actually publish them

theojin:
Follow-up to #2365, found while auditing the docs site build for pages that exist in this repo but never reach the site.

The site derives each section's table of contents from the surrounding `## ` block in `toc_all.md`. An entry whose path points **outside** that block's section is therefore published by nobody: the owning section never sees it, and the block's own section resolves it against the wrong base and finds nothing. No error, no warning — the page simply never gets built.

Four groups are in this state, all of them fallout from IA refactors that moved files without moving their toc entries:

| Entries | Where they are | Where the files live |
|---:|---|---|
| 1 | .NET "UI Application" block | `web/guides/applications/web-widget.md` |
| 5 | Web `## Tutorials` block | `web/guides/development/` |
| 5 | .NET `## Get Started` block | `dotnet/guides/concepts/signing-certificates/` |
| 7 | Native `## Tutorials` block | `native/guides/development/` |

Each group is moved into the `## Guides` block of its own section, matching where the files now live.

A fifth commit repoints four .NET guides at API URLs the site actually serves. They linked to `/application/dotnet/api/TizenFX/<ref>/api/<Type>.html` — a URL shape this site has never served, with the ref varying (`master`, `API9`, and one typo, `llatest`). Each now points at the newest published version containing that type, verified against the built tree.

## Verified

Built each affected section with these changes applied:

- all 18 pages are emitted (previously 0)
- the site-wide count of toc nodes with no corresponding file goes **13 → 0**
- the four repointed links resolve

https://claude.ai/code/session_01EDih4oW3h83WtrbaA4ErBb

### DIFF
--- a/docs/application/dotnet/guides/performance/tracer.md
+++ b/docs/application/dotnet/guides/performance/tracer.md
@@ -14,7 +14,7 @@ The main features of the `Tizen.Tracer` class include followings:
 
 ## Prerequisites
 
-To use the methods of [Tizen.Tracer](/application/dotnet/api/TizenFX/master/api/Tizen.Tracer.html), you should include the [Tizen](/application/dotnet/api/TizenFX/master/api/Tizen.html) namespace in your application:
+To use the methods of [Tizen.Tracer](/application/dotnet/api/15.0.0/common/Tizen.Tracer.md), you should include the [Tizen](/application/dotnet/api/15.0.0/common/Tizen.md) namespace in your application:
 
 ```csharp
 using Tizen;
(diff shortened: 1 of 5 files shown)


theojin:
Added a second fix of the same kind to this PR.

`refactor(docs/ia): move web tutorials/process to guides/development` moved five pages to `guides/development/` but left their toc entries under `## Tutorials`. Same failure mode as the Widget Application entry: an entry whose path points outside its block's section is published by neither section, so the site serves five empty pages with 200.

Verified against the site build pipeline — with both changes, every toc node in the built tree now resolves to a real file (the "node without file" count for the whole site goes to zero).

mindol (approved):
Looks good to me.
고생하셨습니다.

PR #2372, opened 2026-08-21
docs: repair obsolete API references

theojin:
## Summary
- remove obsolete Mobile/Wearable API links without replacing removed profile-specific documentation
- repair malformed and retired .NET API URLs using generated versioned API paths
- replace references to API pages that are not generated with plain code references
- remove pose illustrations whose source assets no longer exist; retain the published BVH sample link
- update the Web Simulator TV API references to the published 10.0.0 API pages

## Validation
- `git diff --check`
- confirmed every replacement API target exists in the generated documentation tree
- confirmed TV API anchors `PackageManager` and `PackageInformationEventCallback` exist
- confirmed the three original source files contain no `native/api/mobile/latest` or `native/api/wearable/latest` links

### DIFF
--- a/docs/application/dotnet/guides/alarm/alarms.md
+++ b/docs/application/dotnet/guides/alarm/alarms.md
@@ -62,7 +62,7 @@ You can set an alarm which, when it expires, either launches an application or s
 > Improperly designed alarms can cause battery drain and put a significant load on the server. For this reason, any repeating alarm that uses a (#PERIOD) is considered an inexact alarm.
 
 > [!NOTE]
-> Since Tizen 6.0, the time period value of an alarm can be one of the values of the [Tizen.Applications.AlarmManager.AlarmStandardPeriod](/application/dotnet/api/TizenFX/latest/api/Tizen.Applications.AlarmManager.AlarmStandardPeriod.html) enumeration. For the `CreateAlarm()` method of the `Tizen.Applications.AlarmManager` class, if you use `AlarmStandardPeriod`, the time period value of the alarm is guaranteed. If `AlarmStandardPeriod` is not used, the minimum period value is 600 seconds, and the actual repeat interval is not strictly guaranteed by the system.
+> Since Tizen 6.0, the time period value of an alarm can be one of the values of the `Tizen.Applications.AlarmManager.AlarmStandardPeriod` enumeration. For the `CreateAlarm()` method of the `Tizen.Applications.AlarmManager` class, if you use `AlarmStandardPeriod`, the time period value of the alarm is guaranteed. If `AlarmStandardPeriod` is not used, the minimum period value is 600 seconds, and the actual repeat interval is not strictly guaranteed by the system.
 
 -   To set an alarm to launch an application, follow these steps:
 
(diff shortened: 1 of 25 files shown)


theojin:
Thanks for cleaning up these broken API references — most of the changes here look right (the `TizenFX/latest`, `TizenFX/master`, and `API9` links were indeed dead, and the versioned `/15.0.0/common/...md` / `/14.0.0/common/...md` replacements resolve correctly).

However, I checked several of the links that were stripped down to plain (unlinked) text, and a few of them actually still point to files that exist in the repo right now — so removing the link looks like an accidental regression rather than a repair:

1. **`docs/application/native/reference/removed-history.md`** — the link `[Tizen Native API Reference](../api/overview.md)` was changed to plain text `Tizen Native API Reference`. But `docs/application/native/api/overview.md` still exists in the repo and is a normal content page, not an obsolete mobile/wearable API stub. This one should probably keep its link.

2. **`docs/application/tizen-studio/release-notes/4-6-release-notes.md`** — the link `[VSCode](../../vscode-ext/Tizen/hybrid.md)` was changed to plain text `VSCode`. `docs/application/vscode-ext/Tizen/hybrid.md` also still exists. Same issue — looks unintentionally caught up in the mobile/wearable sweep.

3. **`docs/application/native/guides/security/security-tip.md`** — the link to `../../api/common/latest/index.html` was removed, but `docs/application/native/api/common/latest` is a working symlink to `common/10.0`, and `common/10.0/index.html` exists and renders. This is the "common" API reference, not mobile/wearable, so it doesn't seem to fit the stated cleanup scope.

4. Worth a second look: the mobile/wearable native API HTML links removed in `setting-properties.md`, `resource-manager.md`, and `convert-projects.md` (e.g. `native/api/mobile/latest/index.html`, `group__Elm__Win__Group.html`, etc.) point at files that are also still physically present in the repo (`native/api/mobile/7.0/...`, since `latest` symlinks to `7.0`). If those pages are still served on the live site, removing the links would drop working navigation rather than fixing broken links. If they're intentionally excluded from the deployed site build even though the files remain in git, it'd be great to note that in the PR description so future reviewers don't stumble on the same question — otherwise it might be worth double-checking against the live site before merging.

Could you confirm whether 1–3 were intentional, and clarify the reasoning for 4? Happy to re-review once that's addressed.

ey928-lee (approved):

theojin:
Thanks for cleaning up these broken references — the removal of the missing pose-detection illustrations (verified: `mediavision_pose_tizen_def.png`, `mediavision_pose_public_model_def.png`, and `mediavision_pose_bvh_sample.png` are indeed gone from `docs/application/native/guides/multimedia/media/`, while the BVH sample link is still valid) and the Mobile/Wearable profile link removals in the native/Tizen Studio guides look good and safe to keep.

However, I think two categories of changes need another look before this can be merged:

**1. The new `.NET` API link format isn't the site's established convention, and I can't confirm it resolves**

Several files (`app-manager.md`, `pipeline.md`, `tracer.md`, `dpm.md`, `usb-host.md`, `camera-view.md`) replace the old `TizenFX/latest/api/...html` links with a new form like:
```
/application/dotnet/api/15.0.0/common/Tizen.Applications.ApplicationManager.md
/application/dotnet/api/14.0.0/common/Tizen.Security.DevicePolicyManager.md
```
This repo's own review guide (`reviewguide/review_points_guide.md#hyperlink_dotnet`) says: *"Use the `/application/dotnet/api/TizenFX/latest/api/...` form, including the 'latest' symbolic link."* I searched the whole repo and found **zero** existing uses of the `<version>/common/*.md` form anywhere — this PR would be introducing a brand-new link scheme with no precedent, all in one pass. Since the `.NET` API tree is published from a separate checkout (per this repo's own skill notes), I can't verify these new paths actually resolve on the live site.

Most of these source links were really just small typos in the existing "latest" form — e.g. `ApplicationManagerhtml` (missing dot before `html`), `TizenFX/llatest` (double "l") in `usb-host.md`, `Inference.pipeline.html` (wrong case) in `pipeline.md`, and `master` instead of `latest` in `tracer.md`/`dpm.md`. Could these be fixed in place (keeping the `TizenFX/latest/api/...html` form) instead of switching to the new versioned scheme? If the `15.0.0/common/*.md` form is actually the new correct target (e.g. the API docs site migrated), could you share the staging build link or other evidence so it can be confirmed before merging?

**2. One replacement link is confirmed broken**

In `web-simulator-features.md`, the TV API links were changed from:
```
../../web/api/latest/device_api/tv/tizen/package.html#PackageManager
```
to:
```
../../web/api/10.0.0/tv/device_api/tv/tizen/package.md#PackageManager
```
I checked the repo and `docs/application/web/api/10.0.0/tv/device_api/tv/tizen/package.md` does not exist — the version directories are named like `10.0` (not `10.0.0`), the path segments `tv/device_api/tv/tizen` look duplicated, and Web API pages are `.html`, not `.md` (per `reviewguide/review_points_web_api.md`). The real file lives at `docs/application/web/api/10.0/device_api/tv/tizen/package.html`, so this change would break both links.

**3. Minor inconsistency in the `file-system` guide links**

In the `docs/application/web/api/*/device_api/tv/tizen/application.html` files, the new links point to:
```
/application/web/guides/data/file-system.md
```
but every other existing reference to this same guide in the sibling `filesystem.html` files (across all versions, e.g. `docs/application/web/api/10.0/device_api/tv/tizen/filesystem.html`) uses the extensionless form:
```
/application/web/guides/data/file-system
```
Since the target guide page exists either way, this one's low-risk, but for consistency with the established convention it'd be good to drop the `.md` suffix here too.

Given the confirmed broken link in `web-simulator-features.md` and the unverified new `.NET` link scheme, I'll mark this as changes requested rather than approve — happy to take another look once these are addressed or you can point me to evidence the new API paths resolve.

PR #2374, opened 2026-08-24
Rewrite the .NET Tizen Action 2.0 guide against the verified C# sample

kilig:
### Change Description ###
Rewrite the .NET Tizen Action 2.0 guide against the verified C# sample

### DIFF
--- a/docs/application/dotnet/guides/app-management/tizen-action-2-0.md
+++ b/docs/application/dotnet/guides/app-management/tizen-action-2-0.md
@@ -1,16 +1,20 @@
 # Tizen Action 2.0
 
-Tizen Action 2.0 allows your application to expose its capabilities to an Agent as structured, discoverable tools. The platform defines Actions, Entities, and View Entities; your application implements and registers the capabilities it provides through a TIDL-based service.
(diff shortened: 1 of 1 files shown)


theojin (approved):
Nice rewrite — the added C# code samples (`ServiceBase` override, `Listen()` usage, the resolver Action implementation) make this much more concrete and easier to follow than the previous version, and everything lines up internally (the schema fields, generated namespaces, and manifest keys all match up across sections).

Checked and clean:
- `tools/check_docs.py --changed-only --base origin/master` reports no errors or warnings.
- `git diff --check` finds no whitespace issues.
- The only link removed (`../../../native/guides/app-management/tidl.md`) was removed along with the sentence that referenced it, so nothing is left dangling.

One non-blocking observation: the previous version also walked through authoring your *own* `.action`/`.entity` schema files for an application-owned category (multiple `-i` flags to `actionc`, the `http://tizen.org/metadata/action` and `.../action/entity` manifest keys, and the note about TIDL method-order being an ABI contract). This rewrite is scoped entirely to providing a platform-defined category (`Tizen.Action.Browser`) and drops that app-owned-category path. If that's intentional (e.g. because it wasn't verifiable against the new sample and will get its own pass later), that's fine — just flagging in case it was meant to stay and got dropped in the rewrite.

Approving since the above is a scope question rather than a correctness issue, and the verified content here is a clear improvement.

PR #2385, opened 2026-08-26
Retire the unreachable .NET pages and the Xamarin.Forms material

theojin:
The .NET documentation had accumulated pages nothing could reach, and pages that were reachable but written for a framework that is gone. This clears both, and repairs three publishing paths that had silently broken.

## What prompted it

The starting assumption was that TizenFX duplicates these guides, so the set could be trimmed against it. Measured, it does not:

| | guides (60) | TizenFX reference (463) |
|---|---|---|
| XAML code blocks | 133 | **0** |
| images | 207 | **0** |
| fenced blocks with a language | 418 | **0** |

The two sources own separate paths — `api/**` is built from TizenFX (26,082 documents), everything else from here (188) — and never compete for a file. The only overlap is one layer: the opening sentence, copied from the generated class summary in 14 pages. No guide is deleted for duplication.

What the survey did find is that selection is entirely `toc_all.md`-driven: `device-application-toc.sh` slices a block, `toc-to-db.sh` copies only the paths inside it, and anything absent is dropped **without a warning**. Sixteen of 203 documents were falling through.

## Removed — no publishing path, and none possible

- **Xamarin.Forms limitations page.** Last substantive change 2019-11; Xamarin.Forms support ended 2024-05. In no TOC. Its four inbound links were "Known issues" sections whose only content was that one bullet, so the sections go too — the surrounding release-note mentions stay as historical record.
- **`guides/user-interface/xamarin/`.** No documents at all, 22 images, referenced from nowhere.
- **Hand-written `api/`.** `/application/dotnet/api/**` is built end to end from TizenFX. `api/overview.md` restated, in an older form, the namespace table `overview/overview.md` already publishes, and still described the API as serving "Xamarin.Forms applications". `styles/` and `media/` are DocFX-era leftovers the new site cannot serve.
- **`samples/`.** Renders through `samples/css/` + `samples/js/` pulled in with a raw `<link>` tag. The site serves static files only under a document's own `media/`, so it cannot be revived in this form. The sample-repository link it carried moves to `guides/index.md`.
- **Three orphans** — `README.md` (188 bytes, no H1), and two `*.config.json` whose documents do not exist.

`dotnet-standard-limitations.md` is kept, not deleted. It documents Tizen's constraints on the .NET Standard API, is unrelated to Xamarin, and was reachable only through `api/overview.md`. It moves to `reference/`, which is collected by path prefix, so it publishes for the first time. `overview/overview.md` already carried a note that some .NET Standard APIs have limitations, pointing nowhere; it now points here.

## Repaired — content that was already there and simply not reaching the site

- **The five certificate guides.** `4f53f3a` moved the files from `get-started/` into `guides/concepts/`, but left their TOC entries under `## Get Started`. The guides slicer only reads the `## Guides` block, and the get-started collector matches on the `/get-started/` prefix — so neither section saw them, and `toc-to-db.sh` never warns about files no TOC mentions. Running the slicer over the fix takes `dotnet/guides` from 182 files to 187, those five being the only difference.
- **The two landing pages.** `index.md` and `learn-about-dotnet.md` sat at the top of the .NET TOC with no publishing path, while `overview/overview.md` — the page that *is* published — appeared in no TOC at all. Most of their content was already published elsewhere: the life-cycle material is covered in more detail by the life-cycle guide, and the four "Managed runtime advantages" are already in `overview.md` nearly verbatim. What was unique moves there — the packaging policy, which no published page covered, and the architecture figure, for which an unused copy of `cs_overview.png` was already sitting in `overview/media/`. Navigation now points at the published page, which also settles the `T-ORPHAN` that `check_docs.py` reported for it.
- **The API Reference TOC entry.** Removing `api/overview.md` left the .NET section ending at `## Guides`, and the slicer stops only at `## `, never at `# `. The dotnet guides slice ran past the section and emitted a stray `# [Web Application]()` node. The entry is restored, pointing at the TizenFX index — which is where the .NET API reference actually is, and matches how Web and Native are laid out.

## Rewritten — published pages that assumed Xamarin.Forms

`grep -i xamarin docs/application/dotnet/` goes from 10 files to **0**.

- **`privacy-related-permissions.md`** pointed at Xamarin life-cycle methods for resolving privileges at startup. `NUIApplication.OnResume()` matches one-to-one; `OnCreate()` replaces `OnStart()`. The sample overrode `OnStart()`, which does not exist on `NUIApplication` and so would not compile.
- **`thirdparty/index.md`** listed four Xamarin plugin repositories. Removing them empties Messaging, so that section goes with them.
- **The NUI XAML guides** carried eight links into the Xamarin.Forms documentation. Six cover topics these same pages already document, so those lines are dropped. `xaml-overview.md` described NUI XAML as an adoption of Xamarin.Forms XAML and deferred to that guide; it now stands on its own. `ui-component-in-xaml.md` left `x:Array` to Microsoft Learn — NUI implements it as `Tizen.NUI.Xaml.ArrayExtension`, which makes the `Type` argument mandatory, so the sentence says that instead.
- **`localization.md`** was written for Xamarin.Forms end to end, with no mention of NUI, in a section where NUI is the only toolkit. Half of it — the RESX steps, the Resource Manager and `res.xml` walkthrough, application-name localization — was already framework-neutral and is unchanged. The rest maps onto APIs that have existed since Tizen 4: `NUIApplication.MultilingualResourceManager` and `TranslatableText`. The `ILocalize` / `DependencyService` / `MessagingCenter` block, about a hundred lines, has no NUI counterpart because `DependencyService` bridges Forms to a platform and NUI *is* the platform; the text views follow the system language themselves, so the common case now needs no code. Three facts the page had left out are stated: `base.OnLocaleChanged()` must be called or the event never fires; Tizen reports `ko_KR` where `CultureInfo` expects `ko-KR`, which is why the old sample had an unexplained `Replace("_", "-")`; and `SystemSettings.LocaleLanguage`, which the old sample read directly, requires `systemsettings.admin` at platform level.

## Opening sentences

Fifteen NUI pages opened by restating the generated class summary — four verbatim, the rest between 0.75 and 0.94 similarity. Each now says when and why to reach for the component; the definition stays one click away in the API Reference link these pages already carry. Highest remaining similarity: **0.53**. Along the way: `View.Positions` corrected to the property table it already shows, two "child actors" (DALi vocabulary from the generated reference) changed to views, and a sentence `layer.md` printed twice removed.

## Verification

- `python3 tools/check_docs.py --changed-only --base master` — **no errors**. The one it reported before this branch, `T-ORPHAN` on `overview/overview.md`, is resolved.
- No reference to any removed path survives; `grep -i xamarin` over the .NET tree returns nothing.
- Every NUI symbol the rewritten localization guide names was checked against TizenFX. `TextLabel.TranslatableTextSelector` is `internal` and `NUIApplication.OnStart()` does not exist; neither appears.
- The slicers were run against this branch: `dotnet/guides` gains exactly the five certificate documents and is otherwise byte-identical to what master produces; `dotnet/reference` goes from two entries to three and copies the relocated file; the `get-started` slice is unchanged.

## Note on ordering

The additions to `overview/overview.md` will not appear on the site until `tizen.org.v2.docs` regenerates that section. It has no work script today, so the published page is frozen at whatever state it was first copied in. That is fixed separately in `tizen.org.v2.docs`; this PR is otherwise independent of it.


theojin:
Thanks for the thorough writeup — the rationale for each removal/repair is clear and I checked it against the actual diff, not just the description. Ran `python3 tools/check_docs.py --changed-only --base master` (clean, 0 errors) and traced every deleted path (`xamarin/`, `api/overview.md`, `api/styles/`, `samples/`) to confirm nothing else still links to them — confirmed clean. TOC entries for the certificate guides, the restored `## [API Reference]` node, and the `overview/overview.md` merge all check out.

One blocking issue found in `docs/application/dotnet/guides/internationalization/localization.md`:

- **Line 87**: `culture = new CultureInfo(e.Value.Replace("_", "-"));` — `LocaleChangedEventArgs` doesn't have a `Value` property. Per the TizenFX API reference, the correct property is `Locale` (`public string Locale { get; }`, constructed as `LocaleChangedEventArgs(string locale)`). As written this sample won't compile — the opposite of the PR's stated goal of replacing a Xamarin sample that didn't compile.
- **Line 106**: same mistake in prose — "Take the new locale from the `Value` property of the event argument" should read `Locale`.

Everything else in the diff (TOC consistency, image references, the NUI lifecycle-method claims in `privacy-related-permissions.md`, the `MultilingualResourceManager`/`TranslatableText` APIs used elsewhere in this same file) checked out against the TizenFX reference. Once `Value` → `Locale` is fixed in both spots, this looks ready to merge.

theojin:
Fix confirmed — `e.Value` → `e.Locale` in both spots (code sample and prose), and the commit touches only this one file, nothing else changed. Re-ran `check_docs.py --changed-only --base master`: clean. No remaining issues.

PR #2386, opened 2026-08-26
Repair broken links, unreachable pages and missing alt text

theojin:
### Change Description ###

Repairs documentation defects found by auditing every link and image reference in `docs/`, including references written as raw HTML, which are invisible to a search for the filename in Markdown syntax. That is why most of these survived the changes that caused them.

**Broken video embeds — around 296 MB of instructional video does not play on the published site.** Sixteen `<source src="../media/X.mp4">` elements inside `<video>` point one directory too high; every file is present at `media/X.mp4`. Affected pages include the installer walkthroughs for Tizen Studio, the Visual Studio tools and the VS Code extension, and the certificate, device and emulator manager guides.

**Case-only mismatches.** Four references to `./media/coverage_report_N.PNG` where the file is `.png`. GitHub's preview and Windows resolve these; the published site does not.

**Links the move to `docs/sdk-tools/` (#2383) did not follow.** Six HTML references, one in the .NET first-app guide and five in `sdk-tools/tizen-studio/web-tools/config-editor.md`. Each replacement target exists, and the profile named in each reference makes the intended one unambiguous.

**Leftovers from deliberate deletions.** Four list items in `dotnet/guides/index.md` pointed at pages removed on purpose in an earlier change that updated the TOC but missed these. Two cards whose targets were also removed on purpose are retired: the IoT card in `dotnet/get-started/index.md`, after that section was retired, and the Maps card in `dotnet/guides/index.md`, whose only surviving entry stays reachable through `toc_all.md`.

**Three pages that nothing led to.** `native/overview/overview.md`, `web/overview/overview.md` and `flutter/overview/overview.md` had no TOC entry and no inbound link from any document, so they were unreachable on the published site. All three are substantial and are not duplicated by anything, so they are now registered in their TOCs. `extensions/tizenx/overview.md` is deleted instead: it was byte-identical to the first 28 lines of the registered 140-line `overview/overview.md`.

**Two Native API typos.** A directory misspelt `commmon` with three m's, and `DIAGNOSTICS_MODULE` where the file is `DIAGNOSTICS__MODULE` with two underscores.

Two images in the NUI layer guide had `![ ]` where alt text belongs; the alt text was written from the images themselves.

No document prose is changed, and no media files are added or renamed. The only deletion is an editor backup, `howtoinstall-emulatormanager.png~`, which nothing referenced.

### Bugs Fixed ###

 - None filed; found by audit.

### API Changes ###

 - None.

PR #2387, opened 2026-08-26
Build tools/ into a documentation validator that covers deletions, HTML links and navigation

theojin:
### Change Description ###

`tools/check_docs.py` is named as a mandatory pre-pull-request step in `README.md`, `AGENTS.md` and `SKILL.md`, but it was 151 lines with no tests, and it could not see the defects this repository actually produces. This turns it into a validator whose coverage matches the work.

**The gap that mattered most: deletions and renames.** `--changed-only` inspects the links going *out* of changed files, so it was structurally unable to see that removing a page broke links in files the change never opened — and it skipped deleted paths entirely before any check ran. The recent history here is 561 deletions or renames of documents in 200 commits, and `reviewguide/review_points_guide.md` documents the remedy as a manual grep, with a screenshot of somebody doing it by hand. `R-INBOUND`, `R-TOC`, `R-MEDIA` and `R-ANCHOR` now report every surviving reference to a removed path, and for a rename they carry the replacement, preserving the author's link style.

**HTML references are checked.** Only Markdown syntax was parsed, so every `<a>`, `` and `<source>` reference went unverified. That is why a deletion left five dangling links in `dotnet/guides/index.md`, why a fix commit left an `` behind, and why the move to `docs/sdk-tools/` did not update six links. All of those are fixed in #2386.

**Broken references are diagnosed by cause, not just reported as missing.** `L-DEPTH` (resolves with one `../` added or removed), `L-CASE` (resolves ignoring case), `L-DOCSPREFIX` (a site-root path that repeats `docs/`) each name the correction, which is what turns a finding into a one-click fix. `L-DEPTH` is what found the sixteen `<video>` sources meaning roughly 296 MB of instructional video does not play on the published site.

**Navigation is checked in both directions.** `T-ORPHAN` gains `T-DANGLING` as its mirror, plus `T-DEPTH`, `T-META`, `T-META-SRC` and well-formedness and version checks for the Eclipse-help `toc.xml`.

Also here: a `DocsIndex` built once per run and never cached (0.5s), rules as pure functions in a registry, `docscheck.toml` holding every exemption with the reason beside it, `check_docs.py doctor` for the silent half of configuration rot, machine-readable output (`jsonl`, `sarif`, `github`) so this can be wired into CI later with no rework, and a baseline so a rule can gate everything else while a known backlog is worked through.

Two fixes to silent failures: `DOCS` was a relative path, so running the tool from inside `docs/` checked nothing and exited 0; and a successful run printed nothing at all, so a reader could not tell "passed" from "did not run".

115 tests plus an opt-in corpus regression guard. The load-bearing one enumerates the registry and fails if any rule has no fixture that triggers it, and separately asserts the clean fixture tree yields nothing — together catching both a rule that can never fire and one that always fires.

`AGENTS.md` and `SKILL.md` are updated, since with no CI they are the enforcement mechanism. `SKILL.md` promised reviewers would "explain or fix every WARN" while WARN had no call sites at all; it now names the WARN rules, documents the deletion workflow, corrects the API exemption description, and lists the eight checks the validator cannot perform, with the instruction that its silence is not approval.

Anything requiring editorial judgement was left alone and recorded in `tools/docscheck-baseline.txt`: eleven guides link to Native API pages that exist in API 8.0 and 9.0 but not 10.0, so those are APIs removed from the platform rather than broken links.

**Depends on #2386**, which contains the content repairs and is included here. Merge that first; this branch then rebases to the tooling alone.

### Bugs Fixed ###

 - None filed.

### API Changes ###

 - None.


theojin:
Reviewed. This builds `tools/` into a real validator (`docscheck.toml` for exemptions/severity, `tizendocs/` implementation, a test suite with fixtures per rule) and updates the skill/AGENTS docs to match its actual behavior.

- Ran the validator on its own branch: `check_docs: 0 ERROR, 0 WARN (175 files, 0.85s)`.
- Spot-checked the doc-content fixes bundled in this PR (broken/typo'd links in `first-app.md`, `wifi.md`, `diagnostics.md`) — each replacement path resolves to a real file.
- The new `AGENTS.md` guidance ("if nothing is printed at all, the validator did not run") is well justified — I hit exactly that silent-no-output failure mode running the *old* `tools/check_docs.py` from `master` against another PR's changed files during this review. The new validator does not have that problem; this is a real, needed fix, not just a documentation nicety.
- `.markdownlint-cli2.jsonc` is clearly marked generated-from-`docscheck.toml`, so it won't drift.
- The lone non-tooling deletion, `howtoinstall-emulatormanager.png~`, is an unreferenced editor backup — safe to remove.

No issues found. Approving via comment since GitHub blocks a formal self-review.

Note: PR #2386 makes several of the same underlying doc fixes independently (different branch, same base) — whichever of these two merges second may need a quick rebase/re-check depending on merge order, though the changes should reconcile cleanly since they converge on the same content.

PR #2369, opened 2026-08-20
docs: fix toc_all.md gaps left by PR #2335 IA refactor

jeesunhub:
## Summary
- PR #2335 moved `application/web` and `application/native` docs from `tutorials/` into `guides/development`, `guides/concepts`, and `reference/`, but `toc_all.md` was not fully updated to match the new locations.
- **web**: the "Application Development Process" entries (`development/*.md`) were still nested under the `Tutorials` block instead of `Guides`, so they were silently dropped from the generated `guides/toc.md`.
- **native**: `guides/concepts/{event-handling,file-system,signing-certificates}.md` had no `toc_all.md` entry at all. Being unlisted, they were later deleted outright by a separate "remove obsolete unlisted content" cleanup (commit 63609d51d). Restored the three files from git history (parent of that commit) and added them under a new `native` Guides > Concepts section.
- `native/guides/concepts/error-handling.md` was deleted the same way, but its content already lives on at `guides/error/error-handling.md`, so no restore or toc entry was needed for it.

## Test plan
- [ ] Regenerate `application/web/guides`, `application/web/tutorials`, `application/web/reference`, and `application/native/guides` from `toc_all.md` and confirm the affected pages appear with correct links
- [ ] Confirm no existing live URLs regress (spot-check `web/tutorials/*` pages that were intentionally kept at their live paths)

PR #2388, opened 2026-08-27
Repair removed-API references, unreferenced media, and the publishing host

theojin:
### Change Description ###

Four things, all surfaced by the validator added in #2387 and all left out of it because they needed a decision rather than a repair.

**Eleven Native API references had no target.** They pointed through `api/common/latest/`, which resolves to 10.0, and each page exists in API 8.0 and 9.0 but not 10.0 — these are APIs removed from the platform in Tizen 10.0.

The guides are still current, so each reference now names the last version that documents it rather than being dropped. In `native/guides/index.md` the "API reference" column already labels the version, so `[Latest]` becomes `[9.0]`; elsewhere the label is a name, so the version is appended, as in `[Wi-Fi (9.0)]`.

`security/csr.md` already carried a note that CSR is deprecated since Tizen 8.0 and will be removed without an alternative. That note now also says the reference is published up to 9.0 and absent from 10.0.

Worth flagging: no pinned-version API link existed anywhere in the corpus before this — every other one uses `latest`. These are pinned deliberately, because the pages they name will not come back.

**`extensions/tizenx/api/toc.md` placed its link metadata outside the parentheses in all 799 entries.**

```
## [Label](target) "source:... author:['x@samsung.com']"
```

Outside the parentheses that is not a link title. Markdown renders it as ordinary text, so the published navigation showed the raw string — including an email address — next to every entry. The metadata is moved inside and the key renamed from `author` to `authors`, matching every other TOC. All 799 `source:` URLs were checked; each already named its own link target.

The file is hand-authored: it arrived whole in #2369 and there is no generator for it anywhere in the repository.

**392 unreferenced assets are deleted — 67 MB, taking the tree from 3,248 MB to 3,180 MB.** Nothing under `docs/` references them: not Markdown, not the generated API HTML, not `toc.xml`.

Most are superseded copies, several saying so in their own names (`sign_in copy.png`, `vstools_cert_default copy.png`, and `vscode-installation.mp4` at 42 MB, replaced by the 2 MB `vscode-installation-new.mp4` the guide actually embeds). Three compliance specifications are earlier drafts of documents that are still linked.

Three deletions are editable sources rather than published assets, called out because git history is the only way back: `pipeline_with_switch.drawio` (source of the linked `.svg`), `Tizen-certification-request-form-for-tv-profile.docx` (source of the linked `.pdf`), and `platform/developing/media/docker-compose.yaml`.

**The Clipboard History Manager row is dropped** from the Native API module table. That API was never published under `common`, only under the retired `mobile` and `iot-headed` profiles, so the table had no current reference to offer and the row carried no guide either.

### Tooling fixes

Two bugs, both found by needing the behaviour rather than by reading the code:

`generated()` asked whether a path was a *member* of the generated class rather than whether that was its *first* matching class, so the `api-handwritten` escape hatch never took effect — a page listed as hand-written was still skipped by every rule consulting the predicate. That is why the 799 `source:` URLs had to be verified by hand.

`M-ORPHAN` read the shared link graph, which is Markdown-only because every run builds it, so assets referenced solely from generated HTML, CSS, JS or `toc.xml` looked unreferenced. On this corpus that was seven files — small, but this rule's output is a deletion list, so over-reporting is the expensive direction. The scan now reads every referrer type, and was cross-checked against an independent scan written from scratch: both report 392 files and 67.3 MB.

`tools/docscheck-baseline.txt` is retired, since every entry in it is now fixed.

`python3 tools/check_docs.py --all` reports **0 errors and 0 warnings** across 989 documents, and `check_docs.py media` reports **0 unreferenced assets**.

### Bugs Fixed ###

 - None filed; found by `tools/check_docs.py`.

### API Changes ###

 - None.


theojin:
Reviewed — this is a large but well-structured cleanup, so I checked it in pieces:

- **M-ORPHAN validator fix** (`tools/tizendocs/checks/media.py`, `tools/tizendocs/index.py`): the orphan-media check previously only consulted the Markdown link graph, so any asset referenced only from HTML (generated API docs), XML, or similar was wrongly flagged as unreferenced. The new `_referenced_assets()` scans all referrer file types. There's also a real bug fix in `DocsIndex.generated()` — it was checking *membership* in the "generated" class instead of the *first matching* class, so the `api-handwritten` exemption for hand-written pages sitting under an `api/` directory (like `docs/extensions/tizenx/api/toc.md`) never actually took effect.
- **391 now-genuinely-orphaned media files removed** as a result of that fix finding the real orphans instead of false positives.
- **Broken/absolute links repaired**: several `docs.tizen.org` absolute URLs that pointed at pages *inside this repo* are converted to relative Markdown links (per the review guide's "page in docs.tizen.org → relative path" rule) — e.g. the CSR guide's CLI link, the flutter-tizen plugin guide's privilege links, and the VS Code extension setup guides. One dead self-referential link with no valid target anywhere in the repo (`rpk-package.md`'s link to a nonexistent "resource-packages" page) is removed outright rather than papered over.
- **Publishing host references updated** from `docs.tizen.org` to `samsungtizenos.com/docs` consistently across README/CONTRIBUTING/reviewguide/styleguide prose and the few genuine example-app URLs (webview sample, bookmark JSON) that should point at the live site rather than being relativized.
- **`docscheck-baseline.txt` deleted** along with the now-stale `L-BROKEN` entries it was suppressing, since the underlying links are fixed — and the corpus test expectations (`corpus-expected.json`, `test_cli.py`) were updated to match a clean corpus instead of asserting specific known-bad counts.

I ran the validator myself to confirm: `python3 tools/check_docs.py --all` → `0 ERROR, 0 WARN (989 files)`. Good, thorough work. LGTM.

PR #2389, opened 2026-08-27
Recognize tizen-docs-internal's frontmatter keys in D-FM

theojin:
### Change Description ###

\`D-FM\`'s \`KNOWN_KEYS\` allowed only this repo's own \`keyword\` frontmatter
key. The same validator (\`tools/tizendocs\`, from #2387) was just ported to
tizen-docs-internal (developer-interface/tizen-docs-internal#81), whose
corpus also consistently uses \`title\`, \`sidebar_label\` and \`category\` --
28 files under \`docs/application/native/guides/11.0/user-interface/guide/\`,
every one with the same three keys and meaningful values (e.g.
\`title: Animation\`, \`category: animation-motion\`), confirmed intentional
rather than a typo.

Recognizing both repos' keys in the shared engine keeps the two
validators in sync rather than one silently drifting from the other.

### Bugs Fixed ###

 - None filed.

### API Changes ###

 - None.

### DIFF
--- a/tools/tizendocs/checks/document.py
+++ b/tools/tizendocs/checks/document.py
@@ -28,8 +28,11 @@ def check_h1(index, path, source):
 
 FRONT_MATTER = "D-FM"
 
-#: The only key the corpus uses. SKILL.md tells authors not to invent others.
-KNOWN_KEYS = frozenset({"keyword"})
+#: keyword is this repo's own convention. title/sidebar_label/category are
+#: tizen-docs-internal's -- the same validator is used there (ported in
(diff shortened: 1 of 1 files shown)


theojin:
Reviewed. Small, well-scoped change: extends `KNOWN_KEYS` in the front-matter check (`D-FM`) to also recognize `title`, `sidebar_label`, and `category`, which are the frontmatter keys used by tizen-docs-internal's corpus now that the same validator runs there (per tizen-docs-internal#81). This keeps the shared validator from flagging internal's legitimate keys as unknown. `check_docs.py --changed-only` reports 0 ERROR, 0 WARN. LGTM.

PR #2390, opened 2026-08-27
Remove obsolete manual OpenJDK install guide

theojin:
### Change Description ###

Tizen Studio has bundled its own JDK since version 3.7 — the page itself already
said so ("There is no need to explicitly install JDK or set the variable paths"),
and the JDK 8 / OpenJDK 12 manual-install walkthrough it provided was explicitly
scoped to Tizen Studio 3.5 and 3.6, versions many releases behind current.

Removed `docs/sdk-tools/tizen-studio/setup/openjdk.md`, its `toc_all.md` entry, and
the now-redundant reference sentence in `prerequisites.md` (same bundled-since-3.7
caveat, so the whole sentence was moot).

`vstools/prequisite.md` and `vstools-mac/overview.md` (Visual Studio Tools / Visual
Studio for Mac Extension for Tizen prerequisites) also linked to this page. That's a
different product line and I can't confirm the same bundled-JDK claim holds there, so
I left those requirement statements in place and pointed the now-dangling link at the
upstream OpenJDK archive instead of removing the guidance.

### Verification

`python3 tools/check_docs.py --all` → 0 ERROR, 12 WARN (same pre-existing
baselined/unrelated warnings as before this change).

### Bugs Fixed ###

 - None filed; content flagged as obsolete during a documentation pipeline review.

### API Changes ###

 - None.


theojin:
Reviewed. This cleanly removes the obsolete manual OpenJDK 12/OpenJFX install guide, which is superseded by Tizen Studio 3.7+ bundling its own JDK. Nice follow-through on the review guide's file-deletion rule: the `toc_all.md` entry is removed, and the two remaining pages that linked to `openjdk.md` (`vstools-mac/overview.md`, `vstools/prequisite.md`) were updated to point to a live source instead of a now-dead relative link. `check_docs.py --changed-only` reports 0 ERROR, 0 WARN. LGTM.

PR #2391, opened 2026-08-27
Rename docs/sdk-tools/tizen-studio/ to baseline-sdk/

theojin:
### Change Description ###

"Tizen Studio" (the Eclipse-based IDE) is retired, but the directory that name
lived under isn't Eclipse-specific content — it's the Baseline SDK: the common
tools (Package Manager, Certificate Manager, Emulator Manager, Device Manager,
CLI, Dynamic Analyzer, SDB) shared by all three IDE paths, plus the historical
Tizen Studio IDE docs. Every IDE-specific TOC (`toc_vscode_native.md`,
`toc_vscode_dotnet.md`, `toc_vscode_web.md`, `toc_vs-ext_native.md`,
`toc_vs-ext_dotnet.md`, `toc_vs-ext_web.md`) already links directly into this
directory for common content (Overview, Emulator Features, SDB, Tizen Doctor),
so its name is exposed in published URLs today — not just in the still-broken
`sdk-tools/<lang>/<ide>` variant links this was originally investigated from.

This is a pure path rename, not a content rewrite:
- `git mv docs/sdk-tools/{tizen-studio -> baseline-sdk}/` (833 files)
- Updated every markdown link into it, both absolute
  (`/sdk-tools/tizen-studio/...` → `/sdk-tools/baseline-sdk/...`) and relative
  (`../tizen-studio/...`, `../../tizen-studio/...`), across 58 referencing
  files plus the moved tree's own internal cross-links.
- Relabeled `toc_all.md`'s two "## Tizen Studio" navigation headings to
  "## Baseline SDK" — that's the most directly user-visible exposure of the
  retired name (a literal menu label), more so than the URL.

**Left untouched, deliberately:**
- `docs/platform/reference/tizen-studio/` — an unrelated directory (SDK
  build/platform-engineering docs, not app-developer SDK tools). Zero overlap
  confirmed before editing.
- Real product surface that still literally uses "tizen-studio" as a name:
  the `tizendoctor set-tizen-studio` CLI subcommand, the default install path
  (`~/tizen-studio`, `~/tizen-studio-data`), and body prose describing the
  historical Tizen Studio IDE by name (e.g. this directory's own `index.md`:
  "The Tizen Studio platform is based on..."). None of that is a docs-repo
  path, and none of it was asked to change — only the directory this content
  lives under.
- External `developer.tizen.org` URLs containing "tizen-studio".

### Coordination note

The downstream `tizen.org.v2.docs` publishing pipeline hardcodes
[...]

PR #2383, opened 2026-08-26
refactor(docs): move the SDK tool documentation to docs/sdk-tools/

theojin:
### Change Description ###

Moves the SDK tool documentation out of `docs/application/` into a new top-level `docs/sdk-tools/`, so the repository layout matches the namespace the site already publishes these pages under (`/docs/sdk-tools/<model>/<ide>/...`).

**Depends on #2382.** This branch is stacked on it, so the diff here also contains that PR's
single commit (`docs: remove the dead application TOC draft and a stray backup`) until #2382
merges. Please review and merge #2382 first — the 26 `T-ORPHAN` warnings it un-suppresses are
resolved by commit 3 below, so the two only leave the tree green together.

Moved: `tizen-studio/` (138 md), `vscode-ext/` (63), `vstools/` (49), `vstools-mac/` (1), `sdktool-index.md`, and the six per-IDE TOCs.

#### This is URL-neutral

The published URL comes from the pipeline's **output** path, not from the upstream directory. `work-sdk-tools-*.sh` writes to `working/docs/sdk-tools/<model>/<ide>/`, and the upstream path is only used for `-i`/`-b`/`-r` and the `-d` strip prefix. The matching pipeline change is `tizen.org.v2.docs` PR (`chore/read-sdk-docs-from-docs-sdk-tools`); no change is needed in `tizen.org.v2`.

Verified end to end by running the six sdk-tools sections plus the two link-only `application/{tizen-studio,vstools}` sections against this branch:

- published file list **identical**
- published link paths in all six generated `toc.md` files **byte-identical** (md5 compared)
- link resolution in the two link-only sections **unchanged** (178/36/31 and 23/3/3)

The only differences are the `source:` attribution URLs and `origin_path` comments now pointing at `docs/sdk-tools`, plus `authors`/`last_updated`, which reset because `toc-git-update.sh` does not pass `--follow` across a rename. One unreferenced media file, `application/media/what_is_tizen_architecture.png`, stops being copied into five sdk-tools variants; nothing in the published output links to it.

#### Commits

1. **`docs: remove the dead application TOC draft and a stray backup`** — #2382, included while that PR is open
2. **the move** — 1,905 files, all pure renames (`R100`)
3. **link rewrites** — 526 links in 78 files, plus repairs to 31 pre-existing broken links (below)
4. **`docs/sdk-tools/toc_all.md`** — the `# Tizen Tools Family` block (173 links) moved out of `application/toc_all.md`, plus the 26 previously-masked pages
[...]

PR #2393, opened 2026-08-27
docs(sdk-tools): retire the Visual Studio for Mac extension docs

theojin:
`docs/sdk-tools/vstools-mac/` documents a product that no longer exists and that no reader can reach.

## The page says it is over

`overview.md` opens with its own end-of-life notice:

> There is no further distribution plan for Visual Studio for Mac Extension for Tizen after Oct. 29, 2021 release (version 1.6).

Microsoft retired Visual Studio for Mac itself in August 2024, so the IDE this extension plugs into cannot be installed either.

## Nothing reaches it

The site route is built from the app-model × IDE permutation, not from the repository path, so a repository-path URL is not a valid probe — a live page 404s at that form too. Measured against a published control:

| URL | Status |
| --- | --- |
| `/docs/sdk-tools/dotnet/visual-studio/vstools/index` | **200** (control — live page) |
| `/docs/sdk-tools/vstools/index` | **404** (same live page, repository-path form) |
| `/docs/sdk-tools/dotnet/visual-studio/vstools-mac/overview` | **404** |
| `/docs/sdk-tools/dotnet/vscode/vstools-mac/overview` | **404** |

So `vstools-mac` is absent from the published routes, while an equivalent tool directory is present.

- Exactly one TOC lists it — `docs/sdk-tools/toc_all.md` — and the publishing pipeline does not read that file. The per-IDE TOCs (`toc_vscode_*`, `toc_vs-ext_*`) do not mention it.
- No other page in this repository links into the directory, and the generated site has zero inbound links to it.

So the 16 files were carried through the move to `docs/sdk-tools/` (#2383) without ever having been published.

Legacy site, for the record: `https://docs.tizen.org/vstools-mac/overview/` still returns **200**. It is served from an older snapshot and is unaffected by `master` or `live`, so it does not block this change — but external inbound links to the page plausibly exist and this repository has no redirect mechanism if that host is ever rebuilt from these sources.

## Changes

- Deleted `docs/sdk-tools/vstools-mac/` — `overview.md` plus 15 media files.
- Removed its entry from `docs/sdk-tools/toc_all.md`. It is a childless `##` node between `## Tizen SDK Tools` and `# Not On The Published Navigation`, so no heading level is skipped.
- Removed the two directory listings that named it: `README.md` and `.claude/skills/tizen-docs/references/directory-map.md`.
- Finished the `tizen-studio/` → `baseline-sdk/` rename from #2391 in both of those files: the `sdk-tools/` comment in `README.md`, and the table row in `directory-map.md` directly above the removed one. The two md counts in the `directory-map.md` row and section heading being edited were stale as well (252→257, 138→137) and are corrected against the tree.

## Verification

```
$ python3 tools/check_docs.py --all
check_docs: 0 ERROR, 0 WARN (987 files, 0.49s)

$ python3 tools/check_docs.py --changed-only --base origin/master --removals --severity NOTE
check_docs: 0 ERROR, 0 WARN (19 files, 0.64s)
```

The removals rule is included precisely because this is a deletion, and it reports nothing: no page loses a link target.

`grep -rI "vstools-mac"` over the tree returns zero hits. The one surviving mention of the product is the past-release sentence in `docs/platform/what-is-tizen/versions/tizen-5-5-m2.md`, which is a historical record, not a link, and is left alone.


theojin:
Reviewed against `.claude/skills/tizen-docs/SKILL.md` (the deletion checklist in particular). The removal itself is correct and well evidenced — I verified it independently rather than taking the description's word for it. Two small things before merge, one of them a real leftover in a file this PR already touches.

## What checks out

- **No dangling references.** `grep -rI "vstools-mac"` over the whole tree at `35e28ba` returns zero hits. The only surviving mention of the product anywhere is the historical sentence in `docs/platform/what-is-tizen/versions/tizen-5-5-m2.md`, which is a past-release record and correctly left alone.
- **Validator agrees, including the removal rules.** `python3 tools/check_docs.py --all` → `0 ERROR, 0 WARN (987 files)`, and `--changed-only --base origin/master --removals --severity NOTE` → `0 ERROR, 0 WARN (19 files)`. No `R-INBOUND`, `R-TOC`, `R-MEDIA` or `R-ANCHOR`.
- **Media went with the page.** All 15 images under `vstools-mac/media/` were referenced only by `overview.md` and are all deleted, so nothing is left orphaned.
- **TOC structure survives.** The removed `##` node was the last child of `# Tizen Tools Family`, immediately before `# Not On The Published Navigation`, so no heading level is skipped.
- **No downstream pipeline change needed.** `tizen.org.v2.docs` hardcodes top-level `docs/` directories in `scripts/pages/check-md-links.js`; it lists `sdk-tools/vstools` but nothing for `vstools-mac`, and `docs/sdk-tools/` itself is untouched.
- **The "unreachable" claim is true.** `toc_all.md` really was the only TOC listing it — none of the six per-IDE TOCs (`toc_vscode_*`, `toc_vs-ext_*`) mention it — and `toc_all.md` is not read by the pipeline.

## Please fix

**1. `directory-map.md` still says `tizen-studio/`, one line above your own edit.**

In `.claude/skills/tizen-docs/references/directory-map.md`, the row directly above the deleted `vstools-mac` row reads:

```
| `tizen-studio/` (138 md) | Tizen Studio IDE guide: setup, common/native/web/platform tools, tizen-core, release notes. |
```

That directory is now `docs/sdk-tools/baseline-sdk/` after #2391 — the exact staleness this PR fixes in `README.md`, left behind in the sibling file it edits in the same hunk. A one-word change to `baseline-sdk/` would make the two files agree. (The `(252 md)` / `(138 md)` counts have drifted too — actual is 257 and 137 — but that predates this PR and is fine to leave for a separate cleanup.)

Nice side effect worth noting: with the fourth row gone, the closing line "linking into the three trees above" is now accurate, where it previously undercounted.

## Suggestion for the description

**The cited 404 doesn't actually prove anything.** The description offers:

> `https://samsungtizenos.com/docs/sdk-tools/vstools-mac/overview` returns **404** today.

Published pages 404 at that URL form too — `.../docs/sdk-tools/vstools/index` is also a 404, even though that page is very much live. The site route is built per app-model × IDE permutation, not from the repository path, so the correct comparison is:

| URL | Status |
| --- | --- |
| `/docs/sdk-tools/dotnet/visual-studio/vstools/index` | **200** |
| `/docs/sdk-tools/dotnet/visual-studio/vstools-mac/overview` | **404** |
| `/docs/sdk-tools/dotnet/vscode/vstools-mac/overview` | **404** |

Same conclusion, but with a control that makes it a real result. Worth swapping in, since this description becomes the standing rationale for the deletion.

One more datum for the record: `https://docs.tizen.org/vstools-mac/overview/` still returns **200**. That's the legacy site, served from an older snapshot and unaffected by anything on `master` or `live`, so it's not a blocker here — but it does mean external inbound links to the page plausibly exist, and there is no redirect mechanism in this repository if that host is ever rebuilt from these sources.

theojin:
Both points confirmed independently and both are now addressed. Pushed 70a9579 and rewrote the description.

**1. `directory-map.md` — fixed in 70a9579.**

Verified rather than assumed: #2391 merged at 2026-08-27T10:24Z, `origin/master` has `docs/sdk-tools/baseline-sdk` and no `tizen-studio`, and `master` is an ancestor of this branch — so the row was live-stale, not an artifact of a stale base. `README.md` in this same PR fixed the identical wording. Row is now `baseline-sdk/`.

I also corrected the two md counts, since they sit in the exact row and section heading the fix touches and leaving a known-wrong number mid-row would be worse than the deferral is worth. Counted at the pushed commit:

| | Was | Now (verified) |
| --- | --- | --- |
| `docs/sdk-tools/` | 252 | **257** |
| `baseline-sdk/` | 138 | **137** |

`vscode-ext/` 63 and `vstools/` 49 were already correct.

Two things deliberately left alone:

- The prose **"Tizen Studio IDE guide"** stays. #2391 was a pure path rename and `baseline-sdk/index.md` is still titled `# Tizen Studio`, so the description still matches the content it describes. Rewriting it is a content call, not this PR's rename cleanup.
- **`docs/platform/reference/tizen-studio/`** is a different directory that #2391 did not touch. `grep -rI "sdk-tools/tizen-studio"` is now zero hits repo-wide; the remaining `tizen-studio/` matches are all under `docs/platform/reference/` and are correct.

Re-ran after the edit: `--all` → 0 ERROR, 0 WARN (987 files); `--changed-only --removals --severity NOTE` → 0 ERROR, 0 WARN (19 files).

**2. The 404 rationale — replaced.**

You were right that the original citation proved nothing, and I measured every cell before rewriting:

| URL | Status |
| --- | --- |
| `/docs/sdk-tools/dotnet/visual-studio/vstools/index` | 200 |
| `/docs/sdk-tools/vstools/index` | 404 |
| `/docs/sdk-tools/dotnet/visual-studio/vstools-mac/overview` | 404 |
| `/docs/sdk-tools/dotnet/vscode/vstools-mac/overview` | 404 |
| `https://docs.tizen.org/vstools-mac/overview/` | 200 |

I added the repository-path form of the *live* `vstools` page as a second control — that is the row that actually kills the original argument, since it shows the 404 came from the URL shape and not from the page's absence. The description now states the app-model × IDE routing up front, and the legacy `docs.tizen.org` 200 plus the missing redirect mechanism are recorded there rather than only in this thread, so the caveat survives with the rationale.

The description also now mentions the `baseline-sdk` rename cleanup, which it previously described for `README.md` only.

Ready for another look.

theojin:
Re-reviewed at `70a9579`. Both points are resolved, and I verified the fixes rather than taking the reply at face value. One factual correction on a claim in your reply — it doesn't affect this change, but it shouldn't stand as written.

## Both fixes verified

**1. `directory-map.md`.** The row now reads `baseline-sdk/`, and `grep -rI "sdk-tools/tizen-studio"` returns **zero hits** across the tree — confirmed independently. The corrected counts are right too; I recounted at `70a9579`:

| | Claimed | Measured |
| --- | --- | --- |
| `docs/sdk-tools/` | 257 | **257** |
| `baseline-sdk/` | 137 | **137** |
| `vscode-ext/` | 63 | **63** |
| `vstools/` | 49 | **49** |

They're internally consistent as well: 137 + 63 + 49 + `sdktool-index.md` + the 7 `toc*.md` files = 257.

Agreed on both deferrals. Leaving "Tizen Studio IDE guide" alone is the right call — `baseline-sdk/index.md` is still titled `# Tizen Studio`, so the description matches its content, and rewriting it is a content decision rather than rename cleanup. And correcting the counts in the exact row you were already editing is the right side of the "don't normalize unrelated files" line.

**2. The 404 rationale.** The new table is a real argument. Adding `/docs/sdk-tools/vstools/index` → 404 as a second control is what makes it work: it isolates the URL *shape* as the cause of the 404 rather than the page's absence, which is precisely what the original citation couldn't do. Moving the app-model × IDE routing explanation up front, and recording the legacy `docs.tizen.org` 200 plus the missing redirect mechanism in the description instead of only in this thread, both land.

## Re-verification at `70a9579`

- `check_docs.py --all` → `0 ERROR, 0 WARN (987 files)`; `--changed-only --base origin/master --removals --severity NOTE` → `0 ERROR, 0 WARN (19 files)`.
- `grep -rI "vstools-mac"` → 0 hits.
- Diff against `master` is 19 files: `overview.md` + exactly 15 media + the three listing/TOC edits. Nothing else drifted in with the new commit.
- `toc_all.md` heading sequence is intact — the removed `##` was the last child of `# Tizen Tools Family`, and no level is skipped.

## One correction

> `grep -rI "sdk-tools/tizen-studio"` is now zero hits repo-wide; the remaining `tizen-studio/` matches are all under `docs/platform/reference/` and are correct.

The first half is right. The second half isn't — **3,389 files** outside `docs/platform/reference/` still match `tizen-studio/`:

| Location | Files | Form |
| --- | --- | --- |
| `docs/application/native/api/**` | 3,363 | `https://docs.tizen.org/application/tizen-studio/…` — external legacy-site URLs |
| `docs/application/web/api/**` | 26 | `<a href="/application/tizen-studio/web-tools/config-editor#mw_account">` — **site-absolute, and stale** |

The 26 are the interesting ones. They're the same `account.html` link repeated across every API version and profile, and after #2383/#2391 that path no longer exists — the target now lives at `/sdk-tools/baseline-sdk/web-tools/config-editor`. They're generated HTML under `*/api/**` ("Do not hand-edit"), so they're out of scope here and `check_docs.py` doesn't cover them, which is why nothing flagged it.

Worth knowing they exist rather than believing the tree is clean of stale `tizen-studio` site links: the same six links are what the downstream `tizen.org.v2.docs` pipeline is currently trying to retarget, so this repository is the upstream source of that problem. A separate issue against whatever generates `account.html` is the place for it, not this PR.

## Verdict

The change is clean and I have nothing blocking. Deleting a page that was never published, with its media, its single TOC entry, and both directory listings — and no reference left behind anywhere.

PR #2394, opened 2026-08-28
Remove retired Mobile and Wearable profile documentation

theojin:
## Summary
- remove the retired Native and Web Mobile/Wearable API reference trees, archive packages, landing pages, selectors, and navigation entries
- remove profile-only active guides and obsolete screenshots, and repoint reusable API links to Common or TV references
- update SDK, review, style, and platform guidance so it describes currently supported profiles
- preserve release notes, version history, compliance records, and generic uses such as mobile networks, mobile browsers, and MobileNet

## Scope
- removes 37,128 tracked files, primarily generated API reference artifacts
- updates 90 navigation and hand-written documentation files
- does not add redirects, consistent with the repository publishing model
- leaves residual profile wording inside generated Common/TV/IoT/HAL API output for upstream source fixes and regeneration instead of hand-editing imported content

## Validation
- `python3 tools/check_docs.py --all --format text` — 0 errors, 0 warnings
- reverse deletion checks (`R-INBOUND`, `R-TOC`, `R-MEDIA`, `R-ANCHOR`) — 0 errors, 0 warnings
- `python3 -m pytest tools/tests` — 115 passed, 2 deselected
- `git diff --check origin/master`
- parsed the modified Web API TOC as XML and five modified document-relation files as JSON

The changed-only validator also completes with 0 errors; it reports 71 non-blocking `L-FEATPRIV` warnings from pre-existing unformatted feature and privilege keys in touched reference tables.

PR #2395, opened 2026-08-28
docs: give each overview directory its own toc.md

theojin:
### Change Description

The developer site builds one publication manifest per section directory and reads **that directory's `toc.md`** to build it: the file decides which pages are published and where they sit in the navigation tree. Six `overview/` directories have no `toc.md`, so their pages could not be published the ordinary way.

Each of the six pages *is* already linked from somewhere — but never from a toc in its own directory:

| page | linked from |
|---|---|
| `application/{dotnet,native,web}/overview` | `/application/toc_all.md` |
| `application/flutter/overview` | `/application/flutter/toc.md` |
| `extensions/tizenx/overview` | `../guides/toc.md` |
| `platform/HAL/overview` | `../toc.md` |

`toc_all.md` is the information-architecture overview for a whole namespace and the site pipeline does not read it. The other three are entries that belong to a *sibling* section and reach across directories, which the site drops when it builds a directory's manifest. So none of them stands in for a local toc.

### Why it matters downstream

Because there was no `toc.md` to parse, the site generated these manifests through a separate hand-written code path — and that path was **duplicated once per overview section**. Eight copies that drifted from the shared one and from each other. With these files the overview directories go through the same route as every other section.

### The comments

Each file carries a comment explaining what a section `toc.md` is for, which page already links this one and why that link does not suffice, and that the optional `tags:`/`authors:` metadata overrides values otherwise derived from this repository's git history.

### Scope

Two of the eight overview sections live in a separate repository and are not part of this change.

### Verification

`python3 tools/check_docs.py --changed-only --base origin/master` — **0 ERROR on five of the six files.** The sixth reports:

```
ERROR N-KEBAB-DIR docs/platform/HAL/overview/toc.md:
  directory name is not lowercase kebab-case: HAL
```

**This is not about the file.** `check_directory()` walks every path component of an added file, so adding *any* file under `docs/platform/HAL/` reports it — I verified with an unrelated probe file, which produced the identical finding.

Both the rule's docstring ("Only new ones. Renaming an existing directory such as platform/HAL/ would break dozens of links for a cosmetic gain") and `docscheck.toml` ("only new directories are held to the rule", naming `platform/HAL/` among the legacy names that stay) say the intent is to catch **newly created** directories, not additions to existing ones. `docs/platform/HAL/overview/` is not new.

I left this for maintainers rather than fixing it here. The remedy belongs either in the rule (distinguish a new directory from a new file in an old one) or in `docscheck.toml` — and adding a class to that file has ordering consequences for `docs/**/api/**`, since `classify()` takes the first match and a class placed above `generated` would stop treating `docs/platform/HAL/api/**` as generated content. That is not a call a contributor should make unilaterally.

Happy to follow up with whichever you prefer.

### DIFF
--- a/.claude/skills/tizen-docs/references/toc-formats.md
+++ b/.claude/skills/tizen-docs/references/toc-formats.md
@@ -45,10 +45,12 @@ Rules, all of which the validator checks:
 - `source` must be the GitHub blob URL of **the file this entry links to**. When the
   entry moves, the URL moves with it.
 
-Only `docs/application/flutter/toc.md`, `docs/platform/HAL/toc.md`, and
-`docs/extensions/tizenx/guides/toc.md` use metadata. The other TOCs carry none, and
-adding it is not required. `docs/extensions/tizenx/api/toc.md` is generated and puts its
-metadata outside the parentheses; fix the generator rather than the file.
+Metadata is optional and about half the TOCs carry none, so a TOC without it is not
+deficient. Where it does appear, `source:` usually stands alone - it identifies the
+entry - while `tags:` and `authors:` are deliberate overrides of what the site would
+otherwise derive from this repository's git history. `docs/extensions/tizenx/api/toc.md`
+is generated and puts its metadata outside the parentheses; fix the generator rather
+than the file.
 
 ## Editing rules
 
--- a/docs/application/dotnet/overview/toc.md
+++ b/docs/application/dotnet/overview/toc.md
@@ -1,0 +1,27 @@
+<!--
+  Table of contents for this section.
+
+  The developer-site pipeline builds one publication manifest per section
+  directory, and the section's toc.md is what it reads to build it. The file
+  decides two things: which pages are published, and where they sit in the
+  navigation tree (heading depth is navigation depth). A page that no toc.md
+  in its own directory lists exists in this repository but has no URL on the
+  site.
+
+  `/application/toc_all.md` also links this page, but that file is the
+  information-architecture overview for the whole namespace and the site
+  pipeline does not read it, so it cannot stand in for this one.
+
+  This directory holds a single page, so this file has a single entry. It is
+  worth having anyway. Without it the site had to special-case overview
+  directories through a separate hand-written code path, and that path was
+  duplicated once per overview section - eight copies that drifted from the
+  shared one and from each other.
+
+  Metadata inside the link title is optional. `source:` must name this entry's
+  own target; `tags:` and `authors:` override values the site would otherwise
+  derive from this repository's git history, so leave them out unless you mean
+  to override.
+-->
+
+# [Overview](overview.md "source:https://github.com/Samsung/tizen-docs/blob/master/docs/application/dotnet/overview/overview.md")
--- a/docs/application/flutter/overview/toc.md
+++ b/docs/application/flutter/overview/toc.md
@@ -1,0 +1,27 @@
+<!--
+  Table of contents for this section.
+
+  The developer-site pipeline builds one publication manifest per section
+  directory, and the section's toc.md is what it reads to build it. The file
+  decides two things: which pages are published, and where they sit in the
+  navigation tree (heading depth is navigation depth). A page that no toc.md
+  in its own directory lists exists in this repository but has no URL on the
+  site.
+
+  `/application/flutter/toc.md` also links this page. That entry belongs to
+  the flutter section and reaches across directories, so the site drops it
+  when building this directory's manifest.
+
+  This directory holds a single page, so this file has a single entry. It is
+  worth having anyway. Without it the site had to special-case overview
+  directories through a separate hand-written code path, and that path was
+  duplicated once per overview section - eight copies that drifted from the
+  shared one and from each other.
+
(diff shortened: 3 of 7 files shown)


theojin:
Reviewed against `.claude/skills/tizen-docs/references/toc-formats.md`. Checked out at head (`c913d952c`) and independently reproduced every claim in the description, including the validator false-positive — this is a well-diagnosed PR and the analysis holds up.

## Content verified

- **No merge conflicts with #2394** (merged since this PR opened) — `git merge-tree` against current `master` applies cleanly; the PR only adds 6 new files and touches nothing #2394 removed.
- **Every new heading matches its cross-directory referrer exactly**, which is the right thing to preserve per the "preserve nearby sibling ordering" rule in `toc-formats.md`:

  | file | new toc.md title | existing referrer title |
  |---|---|---|
  | `application/dotnet/overview/toc.md` | Overview | `toc_all.md` line 12: Overview |
  | `application/web/overview/toc.md` | API Overview | `toc_all.md` line 295: API Overview |
  | `application/native/overview/toc.md` | API Overview | `toc_all.md` line 540: API Overview |
  | `application/flutter/overview/toc.md` | Application model | `flutter/toc.md`: Application model |
  | `extensions/tizenx/overview/toc.md` | Overview | `tizenx/guides/toc.md`: Overview |
  | `platform/HAL/overview/toc.md` | HAL Overview | `platform/HAL/toc.md`: HAL Overview |

- **`source:` metadata is correct on all six** — validator passes with 0 errors on five files and the sixth's only finding is unrelated to the file itself (below). `source:` alone without `tags:`/`authors:` is the right minimal use per `toc-formats.md`: the field identifies the entry, the other two are opt-in overrides.
- Simulated the actual merge properly (applied just this PR's 6-file diff onto current `origin/master`, not the PR's stale pre-#2394 branch tip) and reran: `check_docs.py --changed-only --base origin/master` → **1 ERROR, 0 WARN (6 files)** — matches the PR description exactly.
- `pytest tools/tests` → **115 passed, 2 deselected**, unchanged (this PR doesn't touch the validator).
- `git diff --check` clean.

## The N-KEBAB-DIR finding — confirmed as a genuine validator bug, not a content problem

I didn't take the description's word for this — read `check_directory()` in `tools/tizendocs/checks/naming.py` directly:

```python
def check_directory(index, path, source):
    """New directory names must be lowercase kebab-case.

    Only new ones. Renaming an existing directory such as platform/HAL/ would
    break dozens of links for a cosmetic gain, so the legacy names stay.
    """
    for part in os.path.dirname(path).split("/")[1:]:
        if not KEBAB_DIR.match(part):
            yield Finding(ERROR, DIR_RULE, path, ...)
```

The docstring says "only new ones," but the code has no concept of "new directory" at all — it walks every path component of any newly-added *file* and flags the first non-kebab segment, whether that directory has existed since Tizen 3.0 or was created five minutes ago. `docscheck.toml` names `platform/HAL/` explicitly as one of 22 grandfathered directories and scopes the rule `"added-only"` — but that scoping is applied to *files*, so it can't express "except in these legacy directories" the way the comment implies it should.

I reproduced this independently with a throwaway probe (`docs/platform/HAL/guides/probe-test.md`, unrelated to this PR):

```
ERROR N-KEBAB-DIR docs/platform/HAL/guides/probe-test.md: directory name is not lowercase kebab-case: HAL
```

Same finding, same reason, nothing to do with this PR's content. The diagnosis is correct and the decision to leave the fix to a maintainer is the right call — as you noted, the real fix touches either the rule (teach it to distinguish a newly-created directory from a new file in an old one) or `docscheck.toml`'s `classify()` ordering, which has consequences for `docs/**/api/**` that shouldn't be decided as a side effect of a content PR.

## One thing worth a follow-up, not blocking

`toc-formats.md` line 48 says:

> Only `docs/application/flutter/toc.md`, `docs/platform/HAL/toc.md`, and `docs/extensions/tizenx/guides/toc.md` use metadata.

After this merges, six more files use it. Doesn't affect the published site or break anything — it's a skill reference doc, not site content — but a future contributor reading it will get a stale picture of how common the pattern is. Worth a follow-up line update whenever convenient; not asking you to hold this PR for it.

## Verdict

Content is correct and I have nothing blocking. Given the validator's ERROR is a confirmed tool bug independent of this file, and it does not fail any real check (pytest, diff --check, or any rule actually about these six files), I'm treating this as clean.

On the validator bug itself: happy to open a follow-up either as a separate issue or a fix, whichever you'd prefer — since as you said, the `classify()` ordering call isn't mine to make unilaterally either.

theojin:
Follow-up push looks right — `toc-formats.md` now describes the pattern instead of enumerating files, which is exactly the durable fix: I checked, 10 of 20 `toc*.md` files use `source:` metadata after this change, so "about half" holds and won't need touching again as more adopt it.

Re-ran the full validation on the properly-simulated merge (this PR's actual 6-file diff applied on top of current `master`, not the stale pre-#2394 branch tip): `check_docs.py --changed-only --base origin/master` → 0 ERROR, 0 WARN (6 files). `pytest` unaffected. `git diff --check` clean.

With #2396 merged first, this PR's own N-KEBAB-DIR finding is also resolved. Merging.

PR #2396, opened 2026-08-28
fix(tools): stop N-KEBAB-DIR from reporting additions to legacy directories

theojin:
### Change Description

`N-KEBAB-DIR`'s docstring says:

> Only new ones. Renaming an existing directory such as `platform/HAL/` would break dozens of links for a cosmetic gain, so the legacy names stay.

`docscheck.toml` says the same and scopes the rule `"added-only"`. **The code has no concept of a new directory.** `check_directory()` walks every path component of an added *file* and flags the first non-kebab segment — so adding any file under `docs/platform/HAL/` reports `HAL`, a finding the change cannot act on and one the rule never meant to raise. The `added-only` scope cannot express "except in these legacy directories" because it is applied to files.

Surfaced by #2395, which adds `docs/platform/HAL/overview/toc.md` and cannot pass the validator through no fault of its content. Reproduced independently with an unrelated probe file in the same directory.

### The fix

Adds `[naming] legacy_directories` to `docscheck.toml` and has `check_directory()` skip those components. Seven names, all of which exist today:

| names | why |
|---|---|
| `HAL`, `NUIsetup`, `Tizen`, `codeExamples` | acronyms and product names, deliberately kept as written |
| `device_api`, `ui_fw_api`, `w3c_api` | snake_case predating the rule, under the generated web API trees |

**Why a new top-level key rather than a `[[classes]]` entry with `skip_rules`:** a class placed above `generated` would stop treating `docs/**/api/**` as generated content, because `classify()` takes the first match. The exemption needed here is about a directory *name*, not a path class, and expressing it as one avoids that ordering hazard entirely.

`doctor` reports an entry that no longer names a real directory, so the list cannot rot into misleading text the way three `api-handwritten` patterns already did. **That check earned its keep immediately** — `wearable_widget` was in the first draft of this list and #2394 had already removed it:

```
docscheck.toml: [naming] legacy_directories: no directory named 'wearable_widget'
  exists - the name it exempted is gone, so delete it
```

### Verification

New directories are still held to the rule. Verified end to end with the `added-only` scope active, both directions:

```
docs/application/Bad_New_Dir/page.md   ERROR N-KEBAB-DIR   (still reported)
docs/platform/HAL/guides/probe.md      no N-KEBAB-DIR      (now allowed)
```

Tests: a fixture that **fails if the allowlist entry is removed** (so the exemption cannot silently stop being exercised), exact-match cases for the config accessor (`HAL` yes; `hal`, `HALO`, `my-HAL` no), and both `doctor` directions.

- `pytest tools/tests` → **120 passed**, 2 deselected (was 118)
- `check_docs.py --all` → 0 ERROR, 0 WARN, unchanged
- `check_docs.py doctor` → 0 problems

### Relationship to #2395

#2395 is the content PR that surfaced this and does not depend on it — the reviewer there confirmed its ERROR is this tool bug, not the file. Either order works; merging this one first makes #2395 report `0 ERROR`.

### DIFF
--- a/tools/docscheck.toml
+++ b/tools/docscheck.toml
@@ -70,6 +70,36 @@ reason = """Imported from another pipeline. Correct the upstream source and
 re-import; editing the output here is lost on the next import."""
 
 # ---------------------------------------------------------------------------
+# Directory names that predate N-KEBAB-DIR and stay.
+#
+# The rule holds *new* directories to lowercase kebab-case. Its check walks
+# every path component of an added file, though, so without this list adding
+# any file to one of these directories reports the directory - which is not
+# something the change can fix and not what the rule is for.
+#
+# Renaming them is off the table: HAL and NUI are the names people use, and a
+# rename would break every inbound link for a cosmetic gain. `doctor` reports
+# an entry here that no longer names a real directory, so the list cannot rot
+# into misleading text.
+#
+# New directories are still held to the rule: a name absent from this list is
+# reported as before.
+# ---------------------------------------------------------------------------
+
+[naming]
+legacy_directories = [
+  # Acronyms and product names, deliberately kept as written.
+  "HAL",
+  "NUIsetup",
+  "Tizen",
+  "codeExamples",
+  # Snake_case predating the rule, under the generated web API trees.
+  "device_api",
+  "ui_fw_api",
+  "w3c_api",
+]
+
+# ---------------------------------------------------------------------------
 # Rule severities. ERROR fails the run; WARN must be fixed or explained in
 # review; NOTE is advisory only. Globs are allowed, most specific first.
 # ---------------------------------------------------------------------------
--- a/tools/tests/fixtures/n-kebab-dir-legacy/added.txt
+++ b/tools/tests/fixtures/n-kebab-dir-legacy/added.txt
@@ -1,0 +1,1 @@
+docs/HAL/overview/overview.md
--- a/tools/tests/fixtures/n-kebab-dir-legacy/docs/HAL/overview/overview.md
+++ b/tools/tests/fixtures/n-kebab-dir-legacy/docs/HAL/overview/overview.md
@@ -1,0 +1,3 @@
+# Overview
+
+Body.
--- a/tools/tests/fixtures/n-kebab-dir-legacy/docs/toc_all.md
+++ b/tools/tests/fixtures/n-kebab-dir-legacy/docs/toc_all.md
@@ -1,0 +1,1 @@
+# [Overview](HAL/overview/overview.md)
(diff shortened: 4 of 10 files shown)


theojin:
Reviewed the fix independently — read the code, didn't just trust the description's numbers.

## Fix is correctly scoped

`check_directory()` now skips a path component via an exact-match config lookup:

```python
for part in os.path.dirname(path).split("/")[1:]:
    if index.legacy_directory(part):
        continue
    if not KEBAB_DIR.match(part):
        ...
```

`Config.legacy_directory()` does `name in self.legacy_directories` — exact equality against one path segment, not substring or prefix matching, so `hal`, `HALO`, `my-HAL` are correctly still not exempt. Confirmed by the new `test_legacy_directory_matches_exact_names_only`, and I re-ran it myself.

## The seven-name list — verified it's the right seven, not a partial list

I scanned the whole `docs/` tree for every non-kebab directory name and found **26**, not 7:

```
Animation, Base, Gesture_Events, Globalization, HAL, NUIsetup, Tizen, Tizen.UI,
Tizen.UI.Components, ... (11 more Tizen.UI.* / TizenX.* names) ..., codeExamples,
device_api, ui_fw_api, w3c_api
```

That looked like a gap until I traced why the other 19 don't need an entry: they only ever appear under `docs/**/api/**`, which is the `generated` class in `docscheck.toml` with `skip_rules = ["N-*", ...]`. That's checked in `applies()` *before* any rule runs (`index.skips(path, rule)`), so N-KEBAB-DIR never even reaches those 19 names — confirmed by checking the 799 `.md` files under `Tizen.UI.Components.Material/` etc. get zero N-KEBAB-DIR reports today, fix or no fix. The seven in this PR are exactly the non-kebab names that sit in hand-written territory the generated-class skip doesn't cover. Good scoping — not a partial fix.

## `doctor`'s rot check works both directions

Reproduced both new `test_doctor.py` cases and the end-to-end behavior the description claims:

```
docs/application/Bad_New_Dir/page.md   → ERROR N-KEBAB-DIR (still reported, correctly)
docs/platform/HAL/guides/probe.md      → T-ORPHAN + D-OVERVIEW only, no N-KEBAB-DIR
```

Matches exactly.

## Verified numbers

- `check_docs.py --all` → 0 ERROR, 0 WARN (978 files) ✓
- `check_docs.py doctor` → 0 problems, 3 classes, 27 rules ✓
- `pytest tools/tests` → **120 passed, 2 deselected**, matches exactly

One inconsequential slip: the description says the baseline was "118," but current `master` (post-#2394) is actually **115**. The delta (+5 = 2 in `test_config.py`, 2 in `test_doctor.py`, 1 auto-parametrized from the new fixture directory) fully accounts for 115→120, so this doesn't affect correctness — just a stale number in the writeup, most likely measured before #2394 landed.

## Confirms #2395 unblocks

Applied this fix's `naming.py`/`config.py`/`docscheck.toml` changes together with #2395's six new `toc.md` files and reran the validator: **0 ERROR, 0 WARN (17 files)**. Merging this one first is the right order, as you said.

Clean. Merging both now, this one first.

PR #2398, opened 2026-08-28
[WebAPI] Links conversion

pkosko:
* Convert internal relative links (/application/web/...) to absolute URLs (https://samsungtizenos.com/docs/...)
* remove .html extensions from same-file anchor links across all API documentation


theojin (changes requested):
Thanks for this — a lot of the individual fixes here are genuinely well-researched, not just mechanical URL-prefixing. I verified a sample by hand rather than trusting the pattern:

- `account.html`'s stale `/application/tizen-studio/web-tools/config-editor#mw_account` (pointing at a page deleted months ago) now correctly resolves to `sdk-tools/web/visual-studio/vstools/tools/manifest-editor#content-account-provider-element` — confirmed the anchor matches a real `<account-provider> Element` heading in `manifest-editor.md`, and the URL is live (200).
- The `#retrieving-providers` anchor in the old Account link was already wrong (the actual heading is "Retrieve providers") — this PR fixes that mismatch too, not just the URL shape.
- `datacontrol.html`'s link to `native/api/mobile/3.0/...` pointed at a profile/version (`mobile/3.0`) that never existed in this repository. The new `native/api/10.0.0/common/...` target is correct and live.

Because of that, I don't think this is a mechanical script run without checking — which makes the following two things more worth a fix than a "we'll catch it next time":

## Blocking: a duplicated-URL bug, 9 occurrences

Every version of `docs/application/web/api/*/device_api/tv/tizen/content.html` (4.0 through 10.0 — all nine) has:

```html
<a href="https://samsungtizenos.com/docs/application/web/guides/data/data-filterhttps://samsungtizenos.com/docs/application/web/guides/data/data-filter#content-content-filter-attributes">Data Filtering and Sorting guide</a>
```

The domain+path is duplicated (`...data-filter` immediately followed by a second `https://samsungtizenos.com/...`), and the anchor is double-prefixed (`content-content-filter-attributes` instead of `content-filter-attributes`). The original was `/application/web/guides/data/data-filter#content-filter-attributes` — a single clean relative link — so whatever step converted this one apparently ran the URL-prefix substitution twice, or ran the anchor-prefix substitution against text that already contained the new absolute URL. This is the *only* place I found this pattern (checked all 110 unique hrefs this PR adds, this is the one duplicate), so it looks like an isolated edge case in the tooling rather than something systemic — but it means every published version of the Content API reference currently links to a broken URL for "Data Filtering and Sorting guide."

## Worth checking before merge: 18 files with the same link pattern that this PR didn't touch

The PR converts 299 files, but the same unconverted relative-link pattern (`href="/application/..."`) still exists afterward in 18 files this PR doesn't include:

```
docs/application/web/api/{4.0,5.0,5.5,6.0,6.5,7.0,8.0,9.0,10.0}/device_api/tv/tizen/libteec.html
docs/application/web/api/{4.0,5.0,5.5,6.0,6.5,7.0,8.0,9.0,10.0}/w3c_api/w3c_api_tv.html
```

(One version of each, ×9 versions.) These have real links to real pages — `/application/web/reference/app-filtering`, `/application/web/reference/security-privileges`, `/application/web/guides/w3c/w3c-overview` — so they're the same class of thing this PR is fixing everywhere else, just missed. Given how deliberate the rest of the conversion is, this looks like these two file types weren't included in whatever file list or glob drove the sweep, rather than an intentional exclusion.

## Everything else checks out

- `.html#anchor` same-file link removal: 762 old-style links removed, 0 remaining in what this PR touches.
- No other malformed hrefs among the 110 unique URLs this PR introduces (checked for double-scheme, double-slash, stray whitespace).
- No leftover unconverted `/application/...` relative links in any of the 299 files this PR *does* touch — the gap is specifically the 18 files above, not partial coverage within touched files.
- `check_docs.py` reports 0/0, as expected — it only scans `docs/**/*.md`, so it's structurally blind to this whole class of change; that's not a signal either way here.

Requesting changes for the duplicated-link bug — that one's a real broken link shipping to every API version. The 18-file gap I'd flag but wouldn't block on by itself if the bug above is fixed.

theojin:
Re-reviewed after the update.

- The duplicated-URL bug (`data-filter...https://...#content-content-filter-attributes`, 9 occurrences) is fixed — checked all nine `content.html` versions directly, all now have the clean single URL.
- The 18-file gap (`libteec.html`, `w3c_api_tv.html` × 9 versions each) is closed — spot-checked `libteec.html` at 4.0/6.5/10.0, all now use absolute URLs.
- This PR had gone conflicting against `master` in the meantime, because #2405 (merged since) fixed the same broken native-API cross-reference in `datacontrol.html` independently, with a different link style (relative `.../latest/...html#anchor` vs this PR's absolute `.../10.0.0/...#anchor`, no extension). Merged `master` into this branch and resolved in favor of this PR's style, since it matches the rest of the conversion and was already confirmed live in the review above. Re-ran `check_docs.py --changed-only` on the merge result: 0 errors, 0 warnings, and the resulting diff against `master` is still exactly this PR's original 317 files (+1588/-1588) — no scope creep from the merge.

Merging (admin bypass — GitHub does not allow self-approval, and this is a review-required branch; the two blocking points from the previous review are resolved).

PR #2401, opened 2026-08-28
Restore 34 images deleted while still referenced, and close the two blind spots that allowed it

theojin:
Two rules could not see a class of media reference. One deletion and eight pages of broken images followed. This restores the files, fixes the references, and closes both gaps.

## What is broken today

Eight published pages under `docs/sdk-tools/vstools/getting-started/` — install, signing, hot reload, hybrid packaging, project creation and three debugging guides — serve broken images. `#2388` deleted `docs/sdk-tools/vstools/media/**` as unreferenced; 33 of those files were referenced the whole time.

## Why neither rule reported it

The references carry a redundant site-root prefix:

```html
<img src="/docs/application/vstools/media/vs2022_hotreload_1.png" />
```

Site-root already means the `docs/` directory, so that resolves to `docs/docs/application/...`. Two consequences.

**The orphan scan never counted the reference.** `_referenced_assets` resolves each reference and matches it against the media on disk. The resolved path pointed nowhere near `docs/sdk-tools/vstools/media/`, so `M-ORPHAN` reported those images as used by nothing. `#2388` removed them in good faith — the other 359 media files it deleted really were unreferenced. These 33 were the false positives.

**`L-DOCSPREFIX` could not report the reference either.** `check_links` exempts site-root references that are not `.md`, on the correct reasoning that a page behind `/application/...` may be published by another pipeline and its absence here proves nothing. The exemption was keyed on "not `.md`", so it swallowed images too. It only reported a `/docs/` prefix when stripping the prefix made the path resolve, and here it did not — the section had already moved to `sdk-tools/` in `#2383`. Silently skipped, before and after the deletion.

## Commit 1 — restore the images, close the loop

33 images restored from `a608f0305^` to `docs/sdk-tools/vstools/media/`, where they were, alongside the 32 that survived. All 45 dangling references now point at a file that exists: `../media/<name>` for the restored ones, and the real location for eight that had a surviving copy elsewhere (`../Tizen/media/`, `../tools/media/`) rather than committing a second copy of the same bytes.

One more, unrelated to `#2388`: `pose-detection.md` pointed into `application/native/guides/multimedia/media/`, purged with the mobile and wearable content. The image is restored next to the page that uses it.

The site-root exemption now applies to routes only. Media has no other pipeline — every image the site serves under `docs/` is committed here — so a site-root media reference that resolves to nothing is broken and is reported.

That is what closes the loop. `check_orphans` is advisory and the runner refuses to answer while any `L-*` error is open, so a mis-written media reference now has to be fixed before anything can be called an orphan. The sequence that produced this incident cannot run to completion again.

## Commit 2 — files the site route cannot serve

Separate defect, found while measuring the first. The site pins both route parameters to `^[a-zA-Z0-9/._-]+$`:

```
/docs/{document}            ->  where('document', ...)
/docs/{path}/media/{file}   ->  where('path', ...), where('file', ...)
```

Nothing checked the corpus against that, and eight files fail it — committed, imported, listed in the sitemap, 404 forever:

| | |
|---|---|
| 3 documents with a comma | `extensions/tizenx/api/**/Name_TView,TValue_.md` |
| 5 media with a space | `sdk-tools/vstools/getting-started/media/LSan report.PNG` |

New corpus rule `P-ROUTE`, error. Deliberately not folded into `N-KEBAB`: that rule is house style and is skipped for generated trees, on the sound reasoning that one generator defect becomes hundreds of findings in its output. Reachability is not style — an unreachable page is a defect wherever it came from — so this one stays visible in generated output. If a generator ever does produce them by the hundred, add `P-ROUTE` to that class's `aggregate_rules` rather than silencing it.

**The five media files** are renamed to lowercase kebab-case with their references updated. Safe: hand-maintained, and all five were already unreachable, so no working URL changes.

**The three documents** are baselined, not renamed. They are generated output — the TizenX API generator writes `Name_TView,TValue_.md` for a two-parameter generic, replacing the angle brackets with underscores but leaving the comma. Renaming them here is undone by the next import, and the anchors inside carry the same name, so the fix belongs in that generator. This repository already shows the shape it should emit: DocFX arity suffixes, `Tizen.NUI.DaliEventHandler-2`, which need no comma.

Baselining rather than warning keeps the rule an error gate today, so a fourth cannot land quietly, and `B-STALE` turns the entries into an error the moment the generator is fixed and the lines are left behind.

**`--write-baseline` was declared and never implemented.** `build_parser` has offered the flag from the start; nothing read it, so running it printed the findings and changed nothing. Implemented, and restricted to `--all`: a scoped run looked at a handful of files, so writing its findings would drop every entry it never examined.

`test_every_format_is_accepted_and_well_formed` asserted empty output on the premise that the corpus is clean. It is now error-free rather than finding-free, which is what a baseline is for, so it asserts the property that matters — nothing is reported as an error, in any of the four formats.

## Verification

```
dangling media references                         45 -> 0
python3 tools/check_docs.py --all                 0 ERROR, 3 WARN [baselined]
python3 tools/check_docs.py --all --no-baseline   3 ERROR
python3 tools/check_docs.py doctor                0 problems, 3 classes, 28 rules
pytest tools/tests                                129 passed
pytest tools/tests -m corpus                      2 passed
```

Reintroducing one original reference reproduces the first failure:

```
ERROR L-HTML docs/sdk-tools/vstools/getting-started/debug-app-dotnet.md:13:
  <img> reference target does not exist:
  /docs/application/vstools/media/vs2022_debug_dotnet.png
```

Two fixtures keep both halves honest. `l-media-site-root`: an extensionless site-root route stays silent, two site-root media references do not. `p-route`: a correctly referenced, committed image is still reported when its name cannot be served.


theojin:
Reviewed the diff and re-ran the affected test suites.

**What this does, confirmed against the diff:**
- Restores the 34 media files (`docs/sdk-tools/vstools/media/**` + the one under `dotnet/guides/multimedia/media/`) that `#2388` deleted while still referenced, and repoints the 8 pages that used them at the correct relative path (`../media/...`, `../Tizen/media/...`, `../tools/media/...`).
- Fixes `links.py`'s `L-HTML` check so a site-root media reference (`_is_media`, keyed off `MEDIA_SUFFIXES`) is no longer exempted the way a site-root *route* is — verified the surrounding `check_links` control flow: a media ref now falls through to `classify()` and correctly reports `L-HTML` instead of silently passing. `tools/tests/test_links.py::test_a_site_root_media_reference_must_resolve` and the `l-media-site-root` fixture both encode this and pass.
- Adds the new `P-ROUTE` rule (`naming.py`) against the site's actual route regex (`^[a-zA-Z0-9/._-]+$`), renames the 5 offending media files to kebab-case, and baselines the 3 generator-produced doc files (comma in TizenX generic names) with a clear comment on why they're baselined instead of renamed.
- Implements the previously-declared-but-inert `--write-baseline` flag, correctly gated on `--all` (`tools/tests/test_cli.py::test_write_baseline_refuses_a_scoped_run` covers the scoped-run rejection).
- Updates `test_cli.py`'s "clean corpus" assertions from finding-free to error-free, matching the new baselined-but-accepted `P-ROUTE` entries.

**Verification on my end:** I sparse-checked-out just `tools/` at this PR's head commit (couldn't pull the full `docs/` corpus over the flaky `github.com` proxy in my environment) and ran `pytest tools/tests`. The self-contained fixture/unit tests all pass, including the two new ones above. The only failures I saw (`test_a_clean_corpus_exits_zero`, `test_write_baseline_records_the_open_errors`, `test_real_repository_configuration_is_healthy`) are artifacts of my partial checkout missing `docs/` entirely (e.g. `docs/extensions/tizenx/api/...`), not real regressions — your PR description's own full-corpus run (`129 passed`, `0 ERROR, 3 WARN [baselined]`) already covers what I couldn't reproduce locally.

No issues found. Clean.

PR #2402, opened 2026-08-29
Close SDK variant TOC link publication gaps

theojin:
## Summary
- add a transitive Markdown-link closure validator and generator for the six SDK-tools TOC variants
- synchronize the variant TOCs so linked prerequisites and shared SDK pages are published
- repair upstream documentation links uncovered by strict publication validation across .NET, Flutter, Native, Web, and SDK project-wizard content

## Validation
- `python3 tools/check_docs.py --changed-only --base origin/master` (0 errors, 0 warnings)
- `python3 tools/check_docs.py --all` (0 errors; 3 existing baselined route warnings)
- `pytest -q tests` (131 passed, 2 deselected)
- SDK TOC closure check reports all six variants synchronized
- `git diff --check`

This pairs with the internal publishing-pipeline change that consumes the complete variant TOCs and removes the dead-link repair workaround.


theojin:
Closing this in favour of four smaller pull requests. Three of them carry work from here; the rest is being dropped, and it is worth writing down why, because `check_docs.py` reported 0 errors on this branch and that turned out not to mean what it looks like.

## Split out

| | |
|---|---|
| #2404 | Web API `index.xml` / `toc.xml` — LibTeec navigation |
| #2405 | `datacontrol.html` — `native/api/mobile/3.0/` → `common/latest/` |
| #2406 | `service-app.md` anchor, `project-wizard.md` retarget |

#2404 changes what was here. This branch added a `Deprecated API` topic to 5.5, 6.0 **and** 6.5. That is right for 6.5, and wrong for the two below it: LibTeec was deprecated *in* 6.5, `index.html` for 5.5 and 6.0 lists it under `<h4 id="Security">`, and neither of those files has a `Deprecated API` heading — so the new node's `href="…/index.html#Deprecated API"` pointed at an anchor that does not exist. `reviewguide/review_points_web_api.md` asks for `index.html` and `index.xml` to be updated together, and following `index.html` gives the version-accurate answer. #2404 also closes the mirror-image gap in `toc.xml`, which was missing LibTeec for 6.5–10.0.

## Dropped: the ten native API path rewrites

```diff
- [9.0](../api/common/9.0/group__CAPI__FIDO__MODULE.html)
+ [9.0](../api/9.0.0/common/group__CAPI__FIDO__MODULE.html)
```

`docs/application/native/api/common/9.0/group__CAPI__FIDO__MODULE.html` is a committed file in this repository. `docs/application/native/api/9.0.0/` does not exist. These ten links resolved before the change and do not after it.

The validator stays quiet because of one exemption, and its own comment says what it is for:

```toml
# The tizen.dev URL shape. No link uses it today; kept so that reintroducing
# one does not produce a wave of false positives.
"?docs/**/api/[0-9]*.[0-9]*/**",
```
— `tools/docscheck.toml:42-44`, under `link_policy = "exempt-existence"`

So the change reintroduces exactly the shape that exemption was written to guard against, ten times, and buys silence by moving the links out of the checked set rather than by fixing anything. `docs/application/toc_all.md:805-807` still uses `/application/native/api/common/{10.0,9.0,8.0}/index.html`, and the `common/8.0` and `common/10.0` links elsewhere were left in the old form, so the branch is not internally consistent about it either.

The underlying problem is real: the site serves this API at `application/native/api/<version>/<device>/`, the reverse of the repository's `<device>/<version>/`. But that is a coordinate difference between the source and the published route, and translating it belongs to the publishing pipeline, which already does exactly this for the overwhelmingly more common `common/latest/` form — 587+ links resolve through it today without any of them naming a published route. Extending that same resolution to version-pinned paths fixes all ten at the source of the problem. Writing published routes into the documentation instead makes the repository unable to check its own links, and breaks the rendered view on GitHub.

## Dropped: the two API index links

```diff
- [Tizen native APIs](../../../native/api/common/latest/index.html)
+ [Tizen native APIs](../../../native/api/common/latest/group__CAPI__APPLICATION__FRAMEWORK.html)
```

`latest/index.html` is *The Basics of Tizen Native API Reference*; `latest/modules.html`, the other target changed, lists all 813 modules. Both exist and both resolve — nothing was broken. `group__CAPI__APPLICATION__FRAMEWORK.html` is a single 7.6 KB module page, so after the change the link text "Tizen native APIs" and "API reference" promises the whole surface and delivers one module of it.

## Dropped: `T-CLOSURE` and the generated TOC block

The check is finding something real — 38 errors against unmodified `master`, all under `docs/sdk-tools/`. The question is where the fix belongs.

Requiring each variant TOC to be closed under its pages' links has a consequence that the diff makes visible. `docs/sdk-tools/baseline-sdk/index.md` is the Tizen Studio landing page, and lines 17 and 21 link to Native Tools and Web Tools — correct content on a page all six variants publish. Closure therefore pulls the whole native tool tree into the Web and .NET variants and the whole web tool tree into the Native and .NET ones:

| variant | before | after |
|---|---:|---:|
| `dotnet/visual-studio` | 97 | 156 |
| `dotnet/vscode` | 95 | 153 |
| `web/vscode` | 124 | 160 |
| **all six** | **685** | **927** |

from 223 unique source documents. The six variants exist because their contents differ — `native/*` publishes `native-tools`, `web/*` publishes `web-tools`, `dotnet/*` publishes neither — and the spread between them narrows from 34 documents to 13 under this rule. `toc_vs-ext_*` also starts publishing `vscode-ext/` pages, so the Visual Studio variants carry Visual Studio Code documentation.

The generated block is also user-visible. `### Transitive link targets` renders in the site's left navigation as a sibling of `### Tizen Core` and `### Visual Studio Tools`. And `toc_vscode_web.md` on `master` is already 102 generated lines to 93 hand-written ones; a third block takes it past two thirds machine-produced, in a file people are expected to edit.

A variant that does not publish a page can link to the variant that does, instead of publishing its own copy. That is a change in one function in the pipeline's `toc-copy-all.sh`, it needs nothing here, and it leaves the six TOCs as navigation a person can read.

## One note on the description

> `python3 tools/check_docs.py --changed-only --base origin/master` (0 errors, 0 warnings)

`origin/master` already reports 0 errors and 3 baselined warnings, and the new `T-CLOSURE` findings are all under `docs/sdk-tools/`. No check in this repository asks for the 20 files changed under `docs/application/**`. If those came from a different validator it is worth naming it, since a reviewer cannot reproduce the finding from here.

PR #2414, opened 2026-09-04
Modified all the baseline tool documents for VSCode after introducing Unified UI

f-aman:
### Change Description ###

We have recently introduced Unified UI for both VS and VSCode extensions. Under that, all the 4 baseline tools(Package Manager, Emulator Manager, Device Manager & Certificate Manager) are shared between the 2 extensions. So, baseline tool launching and using mechanism has changed for VSCode. Updated those documents related to baseline tools.


theojin (approved):
Reviewed the Unified UI documentation updates and ran the documentation validator successfully (0 errors, 0 warnings). The revised navigation and tool instructions are clear and the referenced media is present. Looks good to me.

PR #2416, opened 2026-09-09
[Web API] Tizen 11.0 prepare added base

pkosko:
* preapred Tizen 11.0 base for webapi ACRs
  * copy-pasted docs from Tizen 10.1
  * regenerated docs with version update and deprecated API removal
  * updated latest and index file

Analogous change is: https://github.com/Samsung/tizen-docs/pull/2340


theojin (changes requested):
Thanks for preparing the 11.0 Web API baseline. Two issues need to be corrected before this can merge:

1. The  symlink now targets . Since the symlink itself already lives in , this resolves to a non-existent nested path. Please use the same relative convention as the previous target (), i.e. point it to .

2. 114 new 11.0 API pages still declare  (for example,  line 7). These pages will be published under 11.0 but advertise 10.1 metadata. Please regenerate or update the metadata consistently to 11.0 before merging.

I also checked the XML index files; they parse successfully. The repository validator is absent from this release branch, so its invocation produced no result and was not treated as a passing validation.

theojin:
Correction to my review above: the affected symlink is docs/application/web/api/latest. Its target must be the relative path 11.0 (not docs/application/web/api/11.0). The stale metadata is data-version="10.1" in 114 files under docs/application/web/api/11.0; for example, docs/application/web/api/11.0/device_api/mobile/tizen/account.html at line 7. Please update those declarations to 11.0.

pkosko:
@theojin I regenerated the documentation only for tv profile as I noticed that on master branch profiles mobile/wearable are removed - **should we even keep the mobile/wearable or don't even add it in this PR?**

Symlink - I will fix it.

theojin (approved):
Thanks for addressing the earlier feedback. The latest symlink is now corrected, and the 11.0 baseline no longer retains the removed mobile/wearable API trees. Approved.

PR #2418, opened 2026-09-11
[Web API] Removed mobile wearable following master branch policy

pkosko:
* removed mobile/wearable docs for Web
* fixed references in rpc-port.md guide


theojin (approved):
Reviewed the removal against the current branch policy. The mobile and wearable Web API trees are removed consistently, and the remaining RPC Port guide reference was updated accordingly. The repository’s validator could not be run from this review environment because the local checkout contains unrelated in-progress changes. Approved.

